PR #14, opened 2025-03-25
Don't create dat file if it doesn't exist

tomneep:
Currently, if you try and run the code with e.g. `fpm run`, it will crash because we don't have any of the input files in version control. However, at the same time it will create an empty file called `inputdata`. This small change just requires that the file we are trying to open actually exists, and so won't create an empty file. 

I then did this for several more of these. I stopped because I got to a point where it started to load a large number of files named things like `xgrid_bk1_076.txt`. We have these files from Chandan, but it might be worth thinking at this stage how exactly we plan to read input data.

### DIFF
--- a/src/subReadInput.f90
+++ b/src/subReadInput.f90
@@ -7,7 +7,7 @@
         CHARACTER(len=160)  :: filename1
 
         !OPEN(60, FILE = 'geometries/2blk/inputdata', FORM = 'formatted')
-        OPEN(60, FILE = 'inputdata', FORM = 'formatted')
+        OPEN(60, file="inputdata", form="formatted", status="old", action="read")
        !READ(60,*) nx, ny, nz,  &
        !           re,      &
        !           itamax, eps1, pcItaMax, &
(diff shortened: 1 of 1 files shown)


mcbarton:
It may be useful to make use of isostat here (search isostat https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vnaf/index.html) . Its an error code depending on what happens when you try and open the file (to see an example of its use here https://stackoverflow.com/questions/17110923/exception-handling-when-the-file-does-not-exist)

warrickball:
As a baseline I think this is a good start and possibly doesn't need anything further. This will crash if the file doesn't exist and should report that the problem is the missing file. E.g. in a small test program, I get
```
$ ./test.x
At line 13 of file test.f90 (unit = 60)
Fortran runtime error: Cannot open file 'inputdata': No such file or directory
...
```
If you're planning to actually handle the error (i.e., there's some useful way of proceeding in the program without that data file), then AFAIK yes, I think `IOSTAT` is the way to go (and not `ERR`, which looks like a goto).

We probably want to change all such statements in one go.

tomneep:
Closing this because we'll end up redoing this with namelists and there isn't really any hope of getting this in a state that can be merged easily.

PR #15, opened 2025-03-25
Remove gotos in main.f90 and subInterfaceDetail.f90

mcbarton:
This PR removes the gotos in main.f90 and subInterfaceDetail.f90 . 

Gotos remain in other subSearch.f90 and subPcorVcor.f90, but these shall be left for now until it is decided what is happening with the commented out code, since how to remove the gotos in these files will depend on what happens with the commented out out code.

### DIFF
--- a/app/main.f90
+++ b/app/main.f90
@@ -77,173 +77,174 @@
         CALL writeOutput1
         coarse_flcnt_check=0
         print*, 'adam'
- 1      CONTINUE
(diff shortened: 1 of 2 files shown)


tomneep (approved):

tomneep:
Hi @mcbarton, I've approved the changed but there are now conflicts that can't be merged. I think it is just a case of adding the module definition.

PR #17, opened 2025-03-26
Terminate nested do loops with multiple end dos instead of continue subsearch.f90

mcbarton:
This PR makes it so we terminate nested do loops with multiple end dos instead of continue subsearch.f90


warrickball:
LGTM

PR #28, opened 2025-03-31
Modules everywhere

tomneep:
Convert all the files containing subroutines into modules and import subroutines only by name


tomneep:
Hi @mcbarton, so for merging this into `dev` you just want me to rename the files to remove the "sub"?

mcbarton:
> Hi @mcbarton, so for merging this into `dev` you just want me to rename the files to remove the "sub"?

@tomneep Yes. I'm assuming it was named that way previously since they contained a subroutines. We don't really need any prefix for our files.

mcbarton (approved):

PR #32, opened 2025-04-03
Remove lfortran from build matrix

### LINKED ISSUE #31: Remove `lfortran` from build matrix
[`lfortran`](https://lfortran.org) is currently "pre-beta" and can't compile our code. It is an interesting project, so we should periodically retry it as progress is made, but for now we should remove it from the CI build matrix.

tomneep:
Closes #31. See issue.


tomneep:
Note that the `gcc` build still won't pass here but a fix for that will come when we move to modules everywhere.

mcbarton (approved):

PR #35, opened 2025-04-08
Lint read input

tomneep:
This addresses all but one of the fortitude linter warnings for `ReadInput.f90`. 

The only one that remains is the `use global`, which is to be somewhat expected at the moment until we can better understand/test the code.

It is worth saying that this changes the output of some local testing I've been running but I think in a good way e.g 

```diff
9,12c9,12
<  dxmin =   1.2500000593718141E-003
<  u0 =  0.13078623037965206
<  dt =   2.6904591996552491E-004
<  freq =  0.92920940794060214
---
>  dxmin =   1.2500000000000000E-003
>  u0 =  0.13078623989560303
>  dt =   2.6904588761086462E-004
>  freq =  0.92920951968456877
```

Note `dxmin`, which is defined as 1.25 in the input file then scaled by `0.001` gives a more sensible result now.

### DIFF
--- a/src/ReadInput.f90
+++ b/src/ReadInput.f90
@@ -1,7 +1,7 @@
 module biocfd_read_input
   use, intrinsic :: iso_fortran_env, only: dp => real64, int64
   use global
-  implicit none (external)
+  implicit none
 
   private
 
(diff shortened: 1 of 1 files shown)


tomneep:
Hmm, `nvfortran` fails which is obviously bad news!

tomneep:
It is the `(external)` part it doesn't like. See discussion [here](https://forums.developer.nvidia.com/t/support-for-implicit-none-external/290470) (also interesting because it talks about NVIDIA moving support towards `flang`). I'll fix this here. Any preference what we should do linting-wise? Should we exclude that rule in `fpm.toml` with a comment that it isn't supported by `nvfortran`?

mcbarton (approved):

PR #40, opened 2025-04-09
Remove StressCal1.f90 and StressCal2.f90 files

mcbarton:
These files are unused by the code, so this PR removes them.


tomneep (approved):
Happy to remove these for now. We can always add them back in if they are needed. For now it removes the need for us to understand what they are doing. Will merge into `dev` now.

PR #42, opened 2025-04-09
Remove unused functions nsMomentum, REDBLACKSOR and REDBLACKSOR_old

mcbarton:
This PR removes the unused functions nsMomentum, REDBLACKSOR and REDBLACKSOR_old


tomneep (approved):
Happy to approve this as it is just deleting unused code. Can always add them back in in the future if it turns out they are needed.

PR #44, opened 2025-04-09
Linting Search.f90

tomneep:
Linting `Search.f90`


tomneep:
There are a quite a few `no-real-suffix` lints that I'm a bit hesitant to fix as they are in accelerated code blocks. Since single-precision operations are generally much faster on the GPU than double precision it might be better to leave as they are for now...

mcbarton:
> There are a quite a few `no-real-suffix` lints that I'm a bit hesitant to fix as they are in accelerated code blocks. Since single-precision operations are generally much faster on the GPU than double precision it might be better to leave as they are for now...

I think this is fine to do as well too. My suggestion is to fix everything you feel comfortable to fix for now, and we can fix the rest of linter issues in this file in future PRs.

mcbarton (approved):

mcbarton (approved):

PR #50, opened 2025-04-10
Apply Werror=unused in ci for gfortran

mcbarton:
This PR makes it so the ci errors out if anything is picked up as unused by the gfortran compiler.

### DIFF
--- a/.github/workflows/setup-fortran.yml
+++ b/.github/workflows/setup-fortran.yml
@@ -42,4 +42,4 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
       - run: |
           # Build with FPM
-          fpm build --flag ${{ matrix.toolchain.FFLAGS }}
+          fpm build
--- a/.github/workflows/setup-fortran.yml
+++ b/.github/workflows/setup-fortran.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         toolchain:
-          - {compiler: gcc, version: 14}
+          - {compiler: gcc, version: 14, FFLAGS: "-Werror=unused"}
           - {compiler: intel, version: '2025.0'}
           - {compiler: intel-classic, version: '2021.10'}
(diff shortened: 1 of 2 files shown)


.github/workflows/setup-fortran.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         toolchain:
-          - {compiler: gcc, version: 14}
+          - {compiler: gcc, version: 14, FFLAGS: "-Werror=unused"}
```
tomneep: I think that if we set `FPM_FFLAGS` then we don't need to set it below. I suppose we could set both here?

```suggestion
          - {compiler: gcc, version: 14, FPM_FFLAGS: "-Werror=unused"}
```

.github/workflows/setup-fortran.yml
```diff
@@ -42,4 +42,4 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
       - run: |
           # Build with FPM
-          fpm build
+          fpm build --flag ${{ matrix.toolchain.FFLAGS }}
```
tomneep: If the above suggestion is applied I don't think we need this here

```suggestion
          fpm build
```

tomneep (commented):
I think we can slightly simplify the changes if we set `FPM_FFLAGS` instead of `FFLAGS`. Other than that looks good!

mcbarton:
Suggestions applied

mcbarton:
Need to revert suggested changes, as realised setting FPM_FFLAGS in the matrix this way, will not set it as an environment variable, but I might be able to do something similar as to what was suggested.

mcbarton:
FPM_FFLAGS is now set in the environment getting its value from the matrix we create in the ci @tomneep (see https://github.com/BioFSILab/Bio-CFD/actions/runs/14378838397/job/40317587277?pr=50#step:5:6) .

tomneep (approved):

PR #52, opened 2025-04-10
Fix pedantic warnings (discluding NavierStokes.f90)

mcbarton:
This PR will fix all warnings raised by the pedantic flag. Once https://github.com/BioFSILab/Bio-CFD/pull/50 is in, I will edit the ci to add pedantic to the list of flags we test.

### DIFF
--- a/src/BoundaryConditions.f90
+++ b/src/BoundaryConditions.f90
@@ -132,7 +132,7 @@ SUBROUTINE velocityBC
          DO n = 1, block(g)%solidCellCount
             i = block(g)%solidIndexPtr(n,1)
             j = block(g)%solidIndexPtr(n,2)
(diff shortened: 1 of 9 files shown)


tomneep (commented):
I think this is a good candidate to split into two PRs. Do you mind trying to separate out the changes to NavierStokes.f90 from everything else?

mcbarton:
> I think this is a good candidate to split into two PRs. Do you mind trying to separate out the changes to NavierStokes.f90 from everything else?

I've reverted the changes to NavierStokes.f90 and will update it in a separate PR. The reason it had many changes in comparison to the others, is because I fixed the non conformant tabs issue using fortify.

mcbarton:
I will have to edit the ci to add the pedantic flag in the second PR.

tomneep (approved):

PR #58, opened 2025-04-10
Initial refactor of InterfaceDetail

tomneep:
This is a first attempt at a bit of refactoring of `InterfaceDetail.f90`. I'm just creating the PR now because I want to see how the compilers handle something I've done, specifically a variable declared as optional.

### DIFF
--- a/src/InterfaceDetail.f90
+++ b/src/InterfaceDetail.f90
@@ -9,12 +9,10 @@ module biocfd_interface_detail
 
   contains
 SUBROUTINE interfaceDetail
-        INTEGER (int64) :: i, j, g,  factor, a_maxx, a_maxy,a_maxz, a_mm,&
-        b_maxx, b_maxy, b_maxz, b_mm, a_max_intf_length, b_max_intf_length,  &
-        a_blk_no, b_blk_no, xx1_p, xx2_p, yy1_p, yy2_p, zz1_p, zz2_p,&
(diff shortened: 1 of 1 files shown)


tomneep:
I think this is fine to merge now. There is more that can be done but this is a good start. I have checked that the output matches perfectly what we had before. There is even some code that can probably eventually be deleted just to ensure a perfect match.

mcbarton (approved):

src/InterfaceDetail.f90
```diff
@@ -75,689 +73,221 @@ SUBROUTINE interfaceDetail
...
+          !> TODO: Check whether the counter really needs to be
+          !> returned. Can probably be determined by the size of
+          !> interface details. It is weirdly one less than the size
+          !> of the interface_details array (dim=2), which seems like
+          !> a potential source of bugs.
+          integer(int64), intent(inout) :: interface_counter
+          ! TN: I don't understand why this works, need to read up on
+          ! it. It is not clear to me whether this will work for all
+          ! of our compilers, online discussion is not clear about
+          ! whether this is really part of the standard
```
warrickball, line 205: Just catching up, but which feature is the one you're unsure about? I've seen the `optional` flag used in other codes I think in basically the same way you've used it here. I.e., by testing using the `present` function and acting appropriately. `optional` is definitely part of the standard.

I've never used the `value` keyword before.

PR #59, opened 2025-04-10
Fix all linter issues with NavierStokes.f90 except C121

mcbarton:
This PR fixes all linter issues with NavierStokes.f90 except C121


tomneep (approved):
I *think* this is all fine so will approve and merge. Definitely a place where some unit tests are needed!

PR #63, opened 2025-04-11
Remove commented code in CoarseInterp.f90

mcbarton:
This PR removes commented code in CoarseInterp.f90

### DIFF
--- a/src/CoarseInterp.f90
+++ b/src/CoarseInterp.f90
@@ -444,7 +444,6 @@ subroutine coarseUpdate
         enddo
         ENDDO
         end subroutine coarseUpdate_newv
-        
         subroutine coarseUpdate_pc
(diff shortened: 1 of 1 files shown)


src/CoarseInterp.f90
```diff
@@ -884,15 +437,14 @@ subroutine coarseUpdate_newv
...
-              !! endif
 
 
         enddo
         enddo
         enddo
         ENDDO
         end subroutine coarseUpdate_newv
-!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!11
+
```
tomneep:
```suggestion

```

tomneep: Might not be clear but the suggestion is to remove the trailing whitespace

tomneep (approved):

Issue #37, opened 2025-04-09
Fortitude all the files


tomneep:
This issue is to keep track of linting each file with fortitude. 
The goal would be to lint the files and find a setup that results in the CI passing.

## List of files

- [x] AllocateArrays.f90 #36 
- [x] BoundaryConditions.f90 #41
- [x] CoarseInterp.f90 #55
- [x] CoefficientMatrix.f90 #56
- [x] FineInterp.f90 #57
- [x] FineInterp_bound.f90 #80
- [x] Forcing.f90 #79 #82
- [x] InitialConditions.f90 #72 
- [x] InterfaceDetail.f90 #51 
- [x] Interpolation.f90 #69 
- [x] LastConditions.f90 #49 
- [x] modGlobal.f90 #67 
- [x] NavierStokes.f90 #59
- [x] PcorVcor.f90 #45 
- [x] ReadInput.f90 #35 
- [x] Search.f90 #44 
- [x] ~StressCal1.f90~ #40
- [x] ~StressCal2.f90~ #40
- [x] WriteOutput_corner1.f90 #43
- [x] main.f90 #81

PR #69, opened 2025-04-11
Lint Interpolation (though not used anywhere)

tomneep:
See #37. Note that it appears this file is unused, but contains some subroutines that we might want to use in the future.


tomneep:
Opened #71 for the issue that the code isn't used anywhere

mcbarton (approved):

PR #94, opened 2025-04-15
Testing more tests for search

tomneep:
Adds a couple more tests for Search.f90. Just creating the PR so we can see the coverage report.


tomneep:
Closing this as gone stale. Should soon be in a situation where we can start testing more.

PR #95, opened 2025-04-15
Convert tabs to spaces in NavierStokes.f90

tomneep:
Follow up to #52. I think that the pedantic flag could be added now (as an error flag).

There should be no none-whitespace changes.

Also, adds the `-pedantic-errors` flag to our `gfortran` build CI.

### DIFF
--- a/.github/workflows/setup-fortran.yml
+++ b/.github/workflows/setup-fortran.yml
@@ -27,7 +27,9 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         toolchain:
-          - {compiler: gcc, version: 14, FPM_FFLAGS: "-Werror=unused --coverage"}
+          - compiler: gcc
+            version: 14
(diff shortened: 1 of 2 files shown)


mcbarton:
> Follow up to #52. I think that the pedantic flag could be added now (as an error flag).
> 
> There should be no none-whitespace changes.

Maybe you can add the pedantic flag to this ci in this PR, and we can see if it passes. If it doesn't this PR can still go in (after reverting the added flag to the ci), and i'll fix any remaining errors.

tomneep:
Probably fine to add this presuming the CI passes (there is no reason it shouldn't), but the gfortran docs mention that we should use this option in conjunction with the `-std` option, so perhaps we should choose a standard we are aiming for?

mcbarton (approved):
I think its ok to leave out the std flag for now, as that will be decided soon anyway.

PR #103, opened 2025-04-28
Remove assignment of unallocated arrays

tomneep:
The `uvflu_avg`, `vwflu_avg`, and `uwflu_avg` are set to 0 in initial conditions but the arrays are never allocated. When building/running with `gfortran`, this produces a segfault. I'm a bit confused what `nvfortran` does such that this doesn't happen, but it is definitely not the intended behaviour. 

These variables aren't used anywhere at all so I'm removing them completely. 

The `SUMWSS` and `SIGNWSS` variables are a similar situation. I'm just commenting these for now pending discussion with the team.

Edit:
I've found another case where this happens, with the `rhs` member variable of Blocks. I've removed it.

### DIFF
--- a/src/InitialConditions.f90
+++ b/src/InitialConditions.f90
@@ -58,12 +58,9 @@ module biocfd_initial_conditions
         block(i)%vflu_rms = 0._dp
         block(i)%wflu_rms = 0._dp
         block(i)%pflu_rms = 0._dp
-        block(i)%uvflu_avg = 0._dp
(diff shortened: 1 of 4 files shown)


warrickball:
LGTM. I also did a bit of grepping around those variables and couldn't see them actually used anywhere.

Along the way, I noticed the `block` components `{u,v,w,p}flu_{avg,rms}` appear to be allocated and initialised to zero but never used.

warrickball (approved):

PR #104, opened 2025-04-30
WIP (DO NOT MERGE): Add tagging_th OpenMP directives and small refactor

tomneep:
Should not be merged before #103. 

This PR makes the mistake of doing more than one thing. They are:

### 1. Add OMP directives to loops in `tagging_th`

Local tests using one iteration spend most the time in the `tagging_th` subroutine. Here I've added OpenMP directives to the most expensive loop. On my laptop (M1 Pro MacBook Pro with 10 cores) compiling using `gfortran` and `-fopenmp` allows me to run the code locally in a *reasonable* time (around 12 minutes for a single iteration)

### 2. Small refactor of most expensive loop of `tagging_th`

The main thrust of this loop is finding the closest elements to some central(?) value for each element in the mesh (to the best of my knowledge). Rather than looping through every element in an explicit loop, we can use the `minloc` intrinsic which is cleaner and empirically slightly faster. Further more, because we are just looking for a minimum, we don't actually need to perform a square root operation on the distance vector as the element with minimum^2 is the same as sqrt(minimum^2). Switching to this approach allows us to remove a bunch of variables and saves us another 2m30s of the time taken to run (total time now around 9m30s).

### Todo
- [ ] The `minloc` intrinsic is faster using OpenMP on my laptop, but I'd like to see what (if any) impact this has when generating code for Nvidia GPUs
- There are a bunch of subroutines that are almost identical to `tagging_th` and could benefit from the same improvements (or better, refactor to avoid repetition). They are
  - [ ] `tagging_th_move`
  - [ ] `selectiveRetagging_th`

### DIFF
--- a/src/Search.f90
+++ b/src/Search.f90
@@ -304,7 +304,7 @@ module biocfd_search
      SUBROUTINE tagging_th
 
         INTEGER(int64) :: g, n, m, i, j, k,  nel2Cen, nel2Pnt, sumNodeId
-        REAL(dp)      :: n1x, n1y, n1z, n2x, n2y, n2z, minDis1, minDis, &
+        REAL(dp)      :: minDis1, minDis, &
                          n2dotn, cent_x, cent_y, cent_z, dis_cen, dis_pnt
 
         CHARACTER(LEN=120) :: filename1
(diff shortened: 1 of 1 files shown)


tomneep:
Sadly this doesn't work when running on the Nvidia GPUs. We run out of memory, which maybe isn't surprising because we are assigning an array per CUDA thread, which is a lot of memory. Back to the drawing board!

PR #105, opened 2025-05-06
dStart1 and dFinish1 are only used in main.f90

tomneep:
A nice easy PR.

Two variables I've spotted `dStart1` and `dFinish1` are only used in `main.f90`, so no need for them to be declared in `modGlobal.f90`.


warrickball (approved):
LGTM

PR #106, opened 2025-05-06
Fix termination of openacc directives

### LINKED ISSUE #102: Failure to compile with `-fopenacc` flag with `gfortran`
It's not happy about using type members in the `firstprivate` OpenACC directive e.g.

```
././src/BoundaryConditions.f90:17:43:

   17 |      !$acc firstprivate (uc, deltat, block(g)%nx)
      |                                           1
Error: Syntax error in OpenMP variable list at (1)
././src/BoundaryConditions.f90:40:23:

   40 |      !$acc end parallel
      |                       1
Error: Unexpected !$ACC END PARALLEL statement at (1)
```

tomneep:
This should allow `gfortran` to build the code with the openacc directives enabled.

Closes #102.

Todo
- [x] Test that this doesn't affect any of the output when running on Baskerville

### DIFF
--- a/src/Search.f90
+++ b/src/Search.f90
@@ -292,7 +292,7 @@ module biocfd_search
            block(g)%cosBeta(n)  = block(g)%cosBeta(n)/lenEl*binor                !direction cosine unit normal along y
            block(g)%cosGamma(n) = block(g)%cosGamma(n)/lenEl*binor               !direction cosine unit normal along z
         ENDDO
-       !$acc end parallel
+       !$acc end parallel loop
 
         print*, 'SurfaceNorm done, inor =', inor
 
@@ -373,7 +373,7 @@ module biocfd_search
          END DO
          END DO
          !$omp end parallel do
-!$acc end parallel
+!$acc end parallel loop
 
 !$acc parallel loop collapse(3) default(present)
            DO k = block(g)%k_startSearch, block(g)%k_endSearch
@@ -391,7 +391,7 @@ module biocfd_search
            END DO
            END DO
            END DO
-!$acc end parallel
+!$acc end parallel loop
 
 
          WRITE(filename1,1) g
@@ -553,7 +553,7 @@ module biocfd_search
          END DO
          END DO
          END DO
-!$acc end parallel
+!$acc end parallel loop
 
 !$acc parallel loop collapse(3) default(present)
            DO k = block(g)%k_startSearch, block(g)%k_endSearch
@@ -572,7 +572,7 @@ module biocfd_search
            END DO
            END DO
            END DO
-!$acc end parallel
+!$acc end parallel loop
 
         block(g)%ibCellCount = 0
          block(g)%solidCellCount = 0
@@ -624,7 +624,7 @@ module biocfd_search
                  END DO
                  END DO
                  END DO
-        !$acc end parallel
+        !$acc end parallel loop
 
         !$acc parallel loop default(present)
         DO n = 1, block(g)%ibCellCount
@@ -635,7 +635,7 @@ module biocfd_search
             block(g)%cell2(i, j, k) = 2
 
          END DO
-        !$acc end parallel
+        !$acc end parallel loop
 
         block(g)%TSCellCount = 0
         tscnt=0
@@ -650,7 +650,7 @@ module biocfd_search
                  END DO
                  END DO
                  END DO
-        !$acc end parallel
+        !$acc end parallel loop
 
         block(g)%TSCellCount = tscnt
         print*, 'TScell count =', block(g)%TSCellCount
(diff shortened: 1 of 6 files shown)


src/Search.f90
```diff
@@ -648,15 +648,14 @@ SUBROUTINE findTScells
...
-        !$acc end parallel
+        !$acc end parallel loop
 
         block(g)%TSCellCount = tscnt
         print*, 'TScell count =', block(g)%TSCellCount
 
         ALLOCATE(block(g)%TSIndexPtr(block(g)%TSCellCount,3))
 
         iPt1 = 0
-        !$acc loop collapse(3) seq
```
tomneep, line 661: I'm fairly confident about everything else in this PR, but this line was a bit confusing. I'm not sure what it was trying to achieve. I need to check deleting this is the right thing to do.

src/Search.f90
```diff
@@ -648,15 +648,14 @@ SUBROUTINE findTScells
...
                  DO k = 0, block(g)%nz+3
                  DO j = 0, block(g)%ny+3
                  DO i = 0, block(g)%nx+3
```
tomneep, line 663: Any ideas about this starting the iteration at element 0?

tomneep:
There is some difference in the outputs running 20 iterations on Baskerville. It is a bit difficult to check as when I try and pipe the output to a file the program seems to stall.

tomneep:
Actually there are some values in the output that change from run to run, so I don't think that the changes in this PR are the issue.

tomneep:
Opened #107 about the changing output

src/Search.f90
warrickball: I presume it was meant to offload the 3D loop that starts immediately after it, so perhaps the fix is actually to *add*
```
!$acc end loop
```
at line 672. I note that this isn't a *parallel* loop and I'm not familiar with OpenACC, so this particular directive might be a no-op. 

But it shouldn't remain as is. It should either start *and* end an `!$acc` block or not have any `!$acc` directives.

src/Search.f90
warrickball: Short answer: it looks suspicious to me.

Long answer: indexing from zero, or actually any number, is allowed in Fortran but you have to explicitly set the bounds of the array. I've only done this in fixed size arrays, like something I once wrote for Legendre polynomials had
```f90
subroutine get_Plm(x, lmax, Plm)
    real(dp), intent(in) :: x
    integer, intent(in) :: lmax
    real(dp), intent(out) :: Plm(0:lmax, 0:lmax)
    ...
```
So the output `Plm` has dimensions `(lmax+1, lmax+1)`, both indexed from zero.

I'm not sure about allocatable arrays but the closest thing I can find for `cell2` is

https://github.com/BioFSILab/Bio-CFD/blob/a3363e38fe4b8302e6c5e411f28165c855737733/src/AllocateArrays.f90#L52

I think it's then passed around via `global`, which I presume means the code knows about the lower bounds. (I just distracted down [a rabbit hole about how lower bounds are passed to procedures](https://fortran-lang.discourse.group/t/allocatable-arrays-and-lower-bounds/1087/21).)

But this looks to me like the compiler should complain, if it's checking bounds, and if it's running anyway it might be retrieving garbage data. All the similar loops seem to be over 2 to `block(g)%n{x,y,z}+1`. For the loop from zero to work, I'd expect the allocation statement to be
```f90
            ALLOCATE(block(i)%cell2(0:nx+3,0:ny+3,0:nz+3))
```
though I'm not sure that's even valid syntax.

warrickball: It might just work if the garbage data is never exactly equal to 2 anyway...

src/Search.f90
tomneep: This needs to be added back (but I don't think it needs an `end loop`). So it turns out that openacc tries to parallelise everything inside a parallel region [citation needed, I'm sure I read this somewhere but can't find where] so you need to explicitly mark any internal loops as sequential if they need to be executed sequentially.

By contrast, OpenMP only parallelises loops you explicitly tell it to [again, citation needed]

tomneep: Actually I was wrong, this isn't inside an OpenACC compute region, so this code will NOT run in parallel anyway

src/Search.f90
tomneep: Opened #118 about this

tomneep:
Tested on Baskerville, no change in the output (beyond the small changes highlighted in #107), and no obvious change in performance when doing 20 iterations. Ready to be reviewed.

mcbarton (approved):
Looks good to me. I noticed a small changes in values in the diff, but this was already noted by Tom, and is at the machine precision level I believe. I think this PR should go in.

Issue #33, opened 2025-04-03
Repetition in `subFineInterp_bound.f90`


tomneep:
Just putting down some thoughts here since I've been looking through this file.

This is a big file ~7000 lines. There is a lot of repetition in this file with only small changes. It looks like a lot of copying and pasting. I think we can make the code much easier to understand by creating some functions or subroutines here. I'm flagging it here because I think it could be reasonably low-hanging fruit.

The general block that we should be able to make into a subroutine/function is: https://github.com/BioFSILab/Bio-CFD/blob/444056b5877ba844674bcda97fa0a1c173c0ffd1/src/subFineInterp_bound.f90#L36-L91

Cycling through `p, u, v, w`

tomneep:
Just a note on this that I'm 99% sure that the `fineUpdate_bd_mv` subroutine is a complete subset of `fineUpdate_bd`, with the only difference that  `fineUpdate_bd` loops over the interfaces 

```f90
DO g=1,intflines
...
END DO
```
and `fineUpdate_bd_mv` has an argument of `g` i.e. we could rewrite `fineUpdate_bd` to just be 

```f90
subroutine fineUpdate_bd
integer :: g
do g=1,intflines
  call fineUpdate_bd_mv(g)
end do
end subroutine
```

Alternatively, since `fineUpdate_bd` is only called once in the code (in the `poissonSolver` subroutine in the `biocfd_pcor_vcor` module, we could potentially just put the loop directly in `poissonSolver`

PR #108, opened 2025-05-07
Remove repeated subroutine in FineInterp_bound

tomneep:
This starts to remove some of the repetition in FineInterp_bound (see #33).

Two things are done:

1. `fineUpdate_bd` just becomes a loop over all blocks and a call to `fineUpdate_bd_mv`, which I believe does the same thing that `fineUpdate_bd` used to do
2. In  `fineUpdate_bd_mv`, stepped iterations are performed such that the first and last elements in each dimension are accessed.

Edit: I also applied step 2 above to the `fineUpdate_pc_bd` and `fineUpdate_newv_bd` subroutines.


tomneep:
I'm not sure that GitHub is handling the diff well here because of the amount of repetition. Perhaps it is better to split the "two things" listed in the PR description into different PRs?

tomneep:
Even after rebasing after #113 the diff is still getting confused. Will look at the best way to break it up next week.

tomneep:
Closing this in favour of other (smaller) PRs such as #114

Issue #91, opened 2025-04-14
Performance tracker


mcbarton:
This issue will keep track of the time taken to do 2 iterations on Baskerville, of the 2 butterfly problem.

tomneep:
Build flags
```bash
export FPM_FC="nvfortran"
export FPM_FFLAGS="-Mr8 -acc -mcmodel=medium -gpu=managed -cuda"
```

## Timings (2 iterations)

| Commit | GPU| Time |
|---------| ----  | -----|
| 2dfd96f5c7b744707c5c9f3a90c30d9345a6d89c | `A100-40GB` | `real	4m19.565s, user	4m2.230s, sys	0m14.747s`


## Timings (20 iterations)

From testing, a lot of time in the first iterations is spent writing output (see the writeOutput1 subroutine #109). We should therefore see that iterations beyond the 2nd are faster.

| Commit | GPU| Time | Notes |
|---------| ----  | -----| ------ |
| 2dfd96f5c7b744707c5c9f3a90c30d9345a6d89c | `A100-40GB` | `real	5m47.892s user	5m21.679s sys	0m22.130s`
| b3d5af90a3460916c4a19491815a9310ce48d9e9 | `A100-40GB` | `00:01:02 walltime` | HDF5 output
| 80b393e586e6092ab2e0a03e10627b8a7678c552 | `A100-40GB` | `00:01:02 walltime` | HDF5 output
| efe966985eab0496538ecb72aefe0cb54ed90dd0 | `A100-40GB` | `00:01:00 walltime` | HDF5 output
| f6d6aef38d774e00d95d2cc69cd974c4e1e7e6d7 | `A100-40GB` | `00:00:57 walltime` | HDF5 output
| 48600d6b5cc910b1b422fb62518d370ad5dc229e | `A100-80GB` | `00:00:56 walltime` | HDF5 output

tomneep:
While working on #112 I learned a number of things which might be useful when it comes to optimising the code.

## `-Minfo=accel,inline`

Adding this flag when using `nvfortran` outputs a lot of information about what the compiler is doing with various regions. The `inline` part might only work when using the `-Minline` flag, see below.

## `-Minline`

I'm not 100% sure whether the compiler will try and do some inlining automatically (such as when doing optimisation). I thought that this was important when working on #112 but in the end I found that the code seemed to work just as well without it. It did highlight some things when combined with the `-Minfo` flag above that could have been the culprit for bad performance.

[...]

PR #110, opened 2025-05-12
Add OMP directives to tagging_th loop

tomneep:
Essentially a cherry pick of 2d7f5fbfc9bbbee8aa21f5fb64d53f73d6f79c72 from my original attempt of this (#104), but with a little bit of clean up. I attempted to use `associate` to make the code a little cleaner, but `nvfortran` didn't like it in the `openacc` block (worked fine with gfortran on my laptop)

I experimented a bit with adding `collapse(2)` or `collapse(3)` to the loop, but didn't see any noticeable speed difference. Similarly, making the next triple-nested loop after this parallel didn't have a big impact.

This is really only for running locally, OpenACC will still be used for the offloading on supported hardware.

Running one iteration of the standard problem (see #91) takes about 9m15 on a MacBook Pro M1 Pro.

### DIFF
--- a/src/Search.f90
+++ b/src/Search.f90
@@ -305,7 +305,7 @@ module biocfd_search
 
         INTEGER(int64) :: g, n, m, i, j, k,  nel2Cen, nel2Pnt, sumNodeId
         REAL(dp)      :: minDis1, minDis, &
-                         n2dotn, cent_x, cent_y, cent_z, dis_cen, dis_pnt
+                         n2dotn, dis_cen, dis_pnt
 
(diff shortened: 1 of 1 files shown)


tomneep:
Hi @mcbarton, I think this is ready to be reviewed. It is essentially just adding OMP directives to one of the heavy loops, so we can run in a reasonable time on our laptops (it would be interesting to see how much faster yours is actually 🙂 ). Aside from adding the directives there is a bit of tidying temporary variables.

mcbarton (approved):

PR #112, opened 2025-05-13
First attempt at removing repetition in Forcing.f90

tomneep:
- Old (unused) interpolation functions in Interpolation.f90 replaced with new ones, following wikipedia notation
- New (private) methods added to Forcing.f90 to perform calculation of interpolated value and derivatives (these use the new interpolation functions)
- Use new `compute_value_and_derivatives` subroutine in `pressureForcing1`

Although this only neatens up `pressureForcing1`, the same changes will be applicable to `velocityForcing1`, `pressureForcingGhost`, `velocityForcingGhost`, `pressureForcingField`, and `velocityForcingField`. I'd propose we update these methods in a future PR so this one can focus on the new functionality. 

There is further repetition that can be targeted in this module but this seems like a good starting point.

#### Todo
- [x] Check performance not degraded - no major change for a single iteration
- [x] Check it runs on Baskerville

### DIFF
--- a/src/Forcing.f90
+++ b/src/Forcing.f90
@@ -27,6 +27,7 @@ SUBROUTINE pressureForcing1
  !$acc           dpdn_e, k, j, i, il, jl, kl, i_x1, i_y1,               &
  !$acc           i_z1,ac_z,ac_y,ac_x,at_y,at_z,ac_x_al,ac_y_al,at_x_al,at_y_al)         &
  !$acc default(present)  &
+ !$acc private(derivatives) &
  !$acc firstprivate (block(g)%nx, block(g)%ny,block(g)%nz)
       DO n = 1, block(g)%ibCellCount
          IF (block(g)%ibSurfId(block(g)%nelp(n))==50) THEN
--- a/src/Forcing.f90
+++ b/src/Forcing.f90
@@ -2,6 +2,7 @@ module biocfd_forcing
   use, intrinsic :: iso_fortran_env, only: dp => real64
   ! allow(use-all) - TODO: Aim to fix this in the future
   use global
+  use biocfd_interpolation, only : linear_interpolation, bilinear_interpolation
   implicit none
 
   private
@@ -13,11 +14,10 @@ module biocfd_forcing
 SUBROUTINE pressureForcing1
 
       INTEGER :: n, k, j, i, il, jl, kl, i_x1, i_y1, i_z1, g
-      REAL (dp) :: n1, pos1_x, pos1_y, pos1_z, pt1, &
-                         aval, bval, cval, p_pos1, sur2nodeDis, dpdn, p_x1, p_x2, &
-                         p_y1, p_y2, p_z1, p_z2, p_x1_z1, p_x2_z1, p_x1_z2, p_x2_z2, &
-                         p_z1_x1, p_z2_x1, p_z1_x2, p_z2_x2, h1, h2, dpdn_e, dpdx_e, dpdy_e, dpdz_e
+      REAL (dp) :: n1, pos1_x, pos1_y, pos1_z, pt1, aval, bval, cval, p_pos1, sur2nodeDis, dpdn, &
+                   dpdn_e
 
+      real(dp) :: derivatives(3)
       dpdn = 0._dp
         DO g=blk_start,nblocks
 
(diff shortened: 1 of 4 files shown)


src/Interpolation.f90
```diff
@@ -1,59 +1,38 @@
...
-        REAL (dp), INTENT(IN) :: l_intp_valx, l_intp_x1, l_intp_x2, l_intp_y1, l_intp_y2
-        REAL (dp), INTENT(OUT) :: l_intp_valy
-        varr=0.
-        varr=(l_intp_valx-l_intp_x1)/(l_intp_x2-l_intp_x1)
-        l_intp_valy=(l_intp_y1*(1._dp-varr))+(l_intp_y2*varr)
+                tmp = matmul(reshape(fvals, [2, 2]), reshape([y2 - y, y - y1], [2, 1]))
+                denom = (x2 - x1) * (y2 - y1)
 
-        end subroutine linearInterp
+                tmp_fout = (matmul(reshape([x2 - x, x - x1], [1, 2]), tmp) / denom)
```
tomneep: It might be better to do "by hand" matrix multiplication since we have small matrices here. Maybe something to return to when we get into the "optimisation" stage of the project in earnest.

src/Forcing.f90
```diff
@@ -13,19 +14,17 @@ module biocfd_forcing
...
 
       INTEGER :: n, k, j, i, il, jl, kl, i_x1, i_y1, i_z1, g
-      REAL (dp) :: n1, pos1_x, pos1_y, pos1_z, pt1, &
-                         aval, bval, cval, p_pos1, sur2nodeDis, dpdn, p_x1, p_x2, &
-                         p_y1, p_y2, p_z1, p_z2, p_x1_z1, p_x2_z1, p_x1_z2, p_x2_z2, &
-                         p_z1_x1, p_z2_x1, p_z1_x2, p_z2_x2, h1, h2, dpdn_e, dpdx_e, dpdy_e, dpdz_e
+      REAL (dp) :: n1, pos1_x, pos1_y, pos1_z, pt1, aval, bval, cval, p_pos1, sur2nodeDis, dpdn, &
+                   dpdn_e
 
+      real(dp) :: derivatives(3)
```
tomneep, line 20: This needs to be added to the `$acc` directive as private

src/Interpolation.f90
```diff
@@ -1,59 +1,38 @@
...
 
+        private
 
-        SUBROUTINE distancePts(d_fine_x1, d_fine_y1, d_coarse_x1, d_coarse_y1, d_val_x1, d_val_y1, &
-                               d_fine_frac, d_coarse_frac)
+        contains
 
+        !> Perform linear interpolation between two points
+        !> See https://en.wikipedia.org/wiki/Linear_interpolation
+        pure function linear_interpolation(x, x1, x0, y1, y0) result(y)
```
tomneep: Revisiting this, I'm not sure why I did `x1` before `x0` (and likewise with `y1` and `y0`), will probably switch these around (and same for `x/y1` and `x/y2` in bilinear interpolation)

tomneep:
Something in here is affecting the GPU performance too much sadly. Need to go through with some plan of what to test and see if we can fix it.

Edit:
Actually, the time for two iterations doesn't seem so bad
```
Required (04:07.825 cputime, 4892208K memory used) - 00:04:16 walltime
```
10 iterations takes (compare with #91)
```
(1) Required (05:17.656 cputime, 4881320K memory used) - 00:05:27 walltime
(2) Required (05:05.182 cputime, 5066032K memory used) - 00:05:17 walltime
```
11 iterations takes
```
(1) Required (07:02.762 cputime, 4886136K memory used) - 00:07:11 walltime
(2) Required (07:08.261 cputime, 4882668K memory used) - 00:07:17 walltime
```
(I don't understand the jump between 10 and 11 iterations)


Some output from running with `NV_ACC_TIME=1` environment variable (11 iterations)

0504ea1f8e62b9d59bd8bb725af5c52f052e48ec (dev branch)
```
slurm-1021413.out:  pressureforcing1  NVIDIA  devicenum=0
slurm-1021413.out-    time(us): 29,874
slurm-1021413.out-    25: compute region reached 22 times
slurm-1021413.out-        25: kernel launched 22 times
slurm-1021413.out-            grid: [432]  block: [128]
slurm-1021413.out-             device time(us): total=29,874 max=3,341 min=766 avg=1,357
slurm-1021413.out-            elapsed time(us): total=30,450 max=3,382 min=780 avg=1,384
slurm-1021413.out-    25: data region reached 44 times
```

 4dbb2e009725b50f0da84912c48fa9298480365f (this branch)
```
slurm-1021478.out:  pressureforcing1  NVIDIA  devicenum=0
slurm-1021478.out-    time(us): 117,050
slurm-1021478.out-    25: compute region reached 22 times
slurm-1021478.out-        25: kernel launched 22 times
slurm-1021478.out-            grid: [216]  block: [128]
slurm-1021478.out-             device time(us): total=117,050 max=10,802 min=2,291 avg=5,320
slurm-1021478.out-            elapsed time(us): total=118,003 max=11,115 min=2,311 avg=5,363
slurm-1021478.out-    25: data region reached 44 times
```

src/Interpolation.f90
tomneep: Switched this to by hand mostly in an effort to get the GPU to inline these functions. In the end I'm not 100% sure that they are now inlined but I think avoiding this reshape mess is nice anyway.

tomneep:
I've been testing various options for this to try and get things working acceptably. What I have here is "performant enough" even though I have to admit I don't fully understand what was causing the problems discussed above.

With the current state of the PR, I get

| Commit | GPU| Time |
|---------| ----  | -----|
| 24a3711c6356151f51fa7ede22905e8afd8f7cb5 | `A100-40GB` | `real	5m37.431s user	5m12.498s sys	0m22.093s`

which compares favourably with #91. I **think** that the key was by-hand inlining the `multi_bilinear_interpolation` function. The reason I found that was through a lot of trial and error and I'll add some comments to the performance tracking thread about some potentially useful things. In the end, I think that we can revisit optimisation in the future, but I'm glad that something using functions doesn't completely kill performance as it should allow us to tidy up the code base quite substantially.

tomneep:
Hmmm turns out I can only add one reviewer, but I'd appreciate if @warrickball and @mcbarton could have a look at this. Any comments are welcome, but in particular it would be good to get your opinion on the order of arguments in the two interpolation functions.

src/Interpolation.f90
tomneep: I've switched these around now

tomneep:
@mcbarton I've switched the order of the arguments to the new functions now. I think this can be reviewed and merged when you are happy with it.

mcbarton (approved):

PR #114, opened 2025-05-20
Remove some by-hand looping in fineUpdate_pc_bd

tomneep:
I'm still struggling to get git to give me a sensible diff of #108 (I suppose because so much is changing). This just removes the repetition from the `fineUpdate_pc_bd` subroutine, I'll try and do the others separately. 

The general point is that there was a "by hand" iteration over the ends of each dimensions e.g.

```f90
i = 1
! do something

i = intfr(g)%counterxp
! do the same thing as above
```

but this can be done in a stepped do loop e.g.

```f90
do i=1, intfr(g)%counterxp, intfr(g)%counterxp-1
```
which achieves the same thing.

Note that there is still a lot more reduction in the code that can be done, e.g. make a step variables that allow us to add another loop to remove the remaining duplication, but this can wait for another PR.


warrickball (approved):
LGTM, though I haven't tried running.

tomneep:
@mcbarton, @warrickball has approved this, would you mind merging when you get a chance?

PR #117, opened 2025-05-28
Navier stokes refactor: step 1

tomneep:
This is the start of the refactor of Navier-Stokes.

## What has changed
- Rather than using e.g. `f_1, f_2, f_3,` etc, use a 1D array (2D in the case of s)
  - `theta_n` -> `theta(n)`
  - `f_n` -> `f(n)`
  - `s_mn` -> `s(m, n)`
  - `ak_n` -> `ak(n)`
- Added pure functions to compute `f`, `s`, and `ak`, and use these in `non_uni_coeff`
- When looping in `non_uni_coeff` directly use e.g. `block(g)%nx` rather than reassigning to `nx_var` (not yet done for allocation of arrays, see next section!)

## What will come next 
Open to discussion where the next changes occur, but I'd propose another PR after this is merged
- Make e.g. `block(g)%ca{1..6}_uu` into a 2D array with one of the dimension 6 in length. This isn't done here because it is slightly more involved - same for e.g. `block(g)%ck{1..6}_uu`
- Two new functions along the lines of `compute_ca(theta)` and `compute_ck(theta)`, which will directly calculate those values (will call existing `calculate` functions)
- Have a single `do g=1, nblocks` loop inside `non_uni_coeff` rather than several

### DIFF
--- a/src/NavierStokes.f90
+++ b/src/NavierStokes.f90
@@ -172,98 +172,39 @@ contains
        tmp_dx3=0.5_dp*(block(g)%deltax(i)+block(g)%deltax(i+1))
        tmp_dx4=0.5_dp*(block(g)%deltax(i+1)+block(g)%deltax(i+2))
 
-       theta_1=tmp_dx2/tmp_dx1
-       theta_2=tmp_dx3/tmp_dx2
-       theta_3=tmp_dx4/tmp_dx3
+       theta = [tmp_dx2/tmp_dx1, tmp_dx3/tmp_dx2, tmp_dx4/tmp_dx3]
 
-       f_1=theta_3
-       f_2=2.0_dp*theta_3+theta_3**2.0_dp
-       f_3=3.0_dp*theta_3+3.0_dp*theta_3**2.0_dp+theta_3**3.0_dp
-       f_4=4.0_dp*theta_3+6.0_dp*theta_3**2.0_dp+4.0_dp*theta_3**3.0_dp+theta_3**4.0_dp
-       f_5=1.0_dp/(theta_1*theta_2)
-       f_6=(2.0_dp*theta_1+1.0_dp)/((theta_1**2.0_dp)*(theta_2**2.0_dp))
-       f_7=(3.0_dp*theta_1**2.0_dp+3.0_dp*theta_1+1.0_dp)/((theta_1**3.0_dp)*(theta_2**3.0_dp))
-        f_8=(4.0_dp*theta_1**3.0_dp+6.0_dp*theta_1**2.0_dp+4.0_dp*theta_1+1.0_dp)/&
-            ((theta_1**4.0_dp)*(theta_2**4.0_dp))
+       f = compute_f(theta)
+       s = compute_s(theta_2, f)
 
-       s51=-1.0_dp/(theta_2**4.0_dp)
-       s52=f_4
-       s53=f_4
-       s54=s51
-       s55=-(f_1/(theta_2**4.0_dp)+f_4/theta_2)
-       s56=(f_4/(theta_2**2.0_dp)-f_2/(theta_2**4.0_dp))
-       s57=-(f_3/(theta_2**4.0_dp)+f_4/(theta_2**3.0_dp))
+       block(g)%ca1_uv(i) = (s(9, 7) * s(10, 1) - s(10, 7) * s(9, 1))
+       block(g)%ca2_uv(i) = (s(9, 7) * s(10, 2) + s(10, 7) * s(9, 3))
+       block(g)%ca3_uv(i) = (s(9, 5) * s(10, 7) - s(10, 5) * s(9, 7))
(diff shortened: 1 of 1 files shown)


tomneep:
I think this at a stage where it can be reviewed and hopefully merged into `dev`. @warrickball, would you mind taking a look at this one?

src/NavierStokes.f90
```diff
@@ -64,276 +67,86 @@ SUBROUTINE non_uni_coeff
...
-       block(g)%ck4_uu(i)=-ak_5
-       block(g)%ck5_uu(i)=ak_6
-       block(g)%ck6_uu(i)=ak_7
+
+       do i=2, block(g)%nx
+       theta = block(g)%deltax(i:i+2) / block(g)%deltax(i-1:i+1)
+       f = compute_f(theta)
+       s = compute_s(theta(2), f)
+
+       block(g)%ca1_uu(i)=(s(9, 7) * s(10, 1) - s(10, 7) * s(9, 1))
```
warrickball: As a minor matter of style in all these lines, we don't need the outer parentheses any more. I.e., I don't see why we shouldn't write
```suggestion
       block(g)%ca1_uu(i)=s(9, 7) * s(10, 1) - s(10, 7) * s(9, 1)
```

src/NavierStokes.f90
```diff
@@ -1558,4 +1190,88 @@ subroutine nsMomentum2order
...
+       s(5, 3) = f(4)
+       s(5, 4) = s(5, 1)
+       s(5, 5) = -(f(1)/(theta_2**4.0_dp)+f(4)/theta_2)
+       s(5, 6) = (f(4)/(theta_2**2.0_dp)-f(2)/(theta_2**4.0_dp))
+       s(5, 7) = -(f(3)/(theta_2**4.0_dp)+f(4)/(theta_2**3.0_dp))
+
+       s(6, 1) = -1.0_dp
+       s(6, 2) = f(8)
+       s(6, 3) = f(8)
+       s(6, 4) = -1.0
```
warrickball:
```suggestion
       s(6, 4) = -1.0_dp
```

src/NavierStokes.f90
```diff
@@ -1558,4 +1190,88 @@ subroutine nsMomentum2order
...
+       real(dp) :: ak(7)
+
+       ak(1) = (1.0_dp+2.0_dp*theta(1))*(theta(3)+theta(3)**2.0_dp)*theta(2)
+       ak(2) = (1.0_dp+theta(1))*(2.0_dp*theta(3)+theta(3)**2.0_dp)
+       ak(3) = (1.0_dp+theta(1))
+       ak(4) = (1.0_dp+theta(1))*((1.0_dp+theta(3))**2.0_dp)
+       ak(5) = ((1.0_dp+theta(1))**2.0_dp)*(theta(3)+theta(3)**2.0_dp)*theta(2)
+       ak(6) = (theta(1)**2.0_dp)*theta(2)*(theta(3)+theta(3)**2.0_dp)
+       ak(7) = (1.0_dp+theta(1))*(theta(3)+theta(3)**2.0_dp)*theta(2)
+end function compute_ak
```
warrickball:
```suggestion
   end function compute_ak
```
(I think. I might be off by a space or two either way. I've added three spaces.)

src/NavierStokes.f90
```diff
@@ -1558,4 +1190,88 @@ subroutine nsMomentum2order
...
        ENDDO
       end subroutine nsMomentum2order
+
+
+    pure function compute_f(theta) result(f)
+        real(dp), intent(in) :: theta(3)
+        real(dp) :: f(8)
+
+        f(1) = theta(3)
+        f(2) = 2.0_dp*theta(3)+theta(3)**2.0_dp
```
warrickball, line 1200: This is probably much more minor than we need to worry about right now but I've been advised, as a broad rule, to use integer types when I intend to raise to an integer power. I.e., in this case
```suggestion
        f(2) = 2.0_dp*theta(3)+theta(3)**2
```
My understanding is that makes it more likely that the compiler can optimise by expanding the power as a series of multiplications, rather than offloading to a potentially expensive general `pow(float ..., float ...)` function.

If you're used to it, here's [an example in Compiler Explorer](https://godbolt.org/z/zW46csrcW), where I've use both an `int` and a `float`, and the right-hand side shows that `gfortran -O2` does the above (`^int` → multiplications, `^float` → `powf`).

warrickball:
Looks good, aside from a few minor comments on things like excessive brackets and using integer vs float exponents.

Also, good plan to break these changes into multiple PRs. :+1:

src/NavierStokes.f90
tomneep: Good point. I've created #119 for that, with the idea that this PR then doesn't change what we had before.

tomneep:
Thanks @warrickball, I've addressed your comments now

tomneep:
I think that this has essentially been approved now by @warrickball, maybe @mcbarton can merge when he has time?

PR #120, opened 2025-06-17
Navier stokes refactor: step 2

tomneep:
Follow on from #117, making e.g. `block(g)%ca{1..6}_uu` into a 2D array with one of the dimension 6 in length and the same for e.g. `block(g)%ck{1..6}_uu`.

I'm going to set the branch used in #117 as the target and see if GitHub correctly redirects this to `dev` once that is merged. Either way, this shouldn't be merged until after #117.

The gains in terms of LOC aren't great, but every little helps. The next step should save a few 10's of lines, but I couldn't immediately get it to work...

### DIFF
--- a/src/NavierStokes.f90
+++ b/src/NavierStokes.f90
@@ -28,42 +28,16 @@ contains
         ny_var=block(g)%ny
         nz_var=block(g)%nz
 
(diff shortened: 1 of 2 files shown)


mcbarton (approved):
Looks good to me. Tested on Baskerville and everything was fine.

PR #121, opened 2025-06-17
Next step of dealing with FineInterp_bound repetition

tomneep:
Removes repetition in `fineUpdate_newv_bd`. 

Try as I might I can't get a sensible diff out of this. I think that it is the repetition itself causing the issues.

Essentially I remove hard-coded repeated loops like

```f90
i=1

DO k=1,intfr(g)%counterzp
DO j=1,intfr(g)%counteryp

! do something

ENDDO
ENDDO

i=intfr(g)%counterxp

DO k=1,intfr(g)%counterzp
DO j=1,intfr(g)%counteryp

! do the same thing again

ENDDO
ENDDO
```

with 

```f90
DO k=1,intfr(g)%counterzp
DO j=1,intfr(g)%counteryp
DO i=1,intfr(g)%counterxp, intfr(g)%counterxp-1

! do something

ENDDO
ENDDO
ENDDO
```

This will be done again for the final subroutine in this file `fineUpdate_bd_mv` in a future PR.

### DIFF
--- a/src/FineInterp_bound.f90
+++ b/src/FineInterp_bound.f90
@@ -259,10 +259,9 @@ SUBROUTINE fineUpdate_bd
            b_blk_no=intfr(g)%b_blk
 
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!uuuuuuu!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-        i=1
-
         DO k=1,intfr(g)%counterzu
         DO j=1,intfr(g)%counteryu
+        DO i=1,intfr(g)%counterxu, intfr(g)%counterxu-1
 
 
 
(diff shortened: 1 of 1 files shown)


warrickball:
I've manually brought the files up in two tabs and "blinked" back and forth to see some of the changes and they look okay.

I doubt this matters but I note you've consistently ordered the looks as `k`, `j` then `i`, which does mean that in some places you'll be changing the order in which things happen. e.g. in the example in the description, I think the truly equivalent loop would be
```f90
DO i=1,intfr(g)%counterxp, intfr(g)%counterxp-1  ! note the loop over i is now first
DO k=1,intfr(g)%counterzp
DO j=1,intfr(g)%counteryp

! do something

ENDDO
ENDDO
ENDDO
```
I leave it to you whether you prefer the consistent order `k`, `j`, `i` versus "smallest loop first" or even something else.

tomneep:
I don't think that it matters in this case, but we can definitely revisit in the future when we start optimising more?

mcbarton (approved):

PR #122, opened 2025-06-23
Remove unused dp member variable

### LINKED ISSUE #86: Remove unused variable `dp` in Blocks
`lfortran` complained that we redefine `dp` in blocks. It looks like it isn't used anywhere so let's just remove it.

https://github.com/BioFSILab/Bio-CFD/blob/da929fdf5181b6be2b7e62adb9915099ce2e3994/src/modGlobal.f90#L125C33-L125C35

tomneep:
Closes #86. Removes an unused member variable that `lfortran` complains about


mcbarton (approved):

warrickball:
>  ... that lfortran complains about

And risks being confused with the `dp` parameter!?

PR #124, opened 2025-07-04
More navier stokes refactor tidy

tomneep:
This tackles a code block where there is a loop over `TSCellCount`. Variables were computed (potentially) multiple times when they can all just be calculated once. This makes it much clearer to see what the code is actually doing.

### DIFF
--- a/src/NavierStokes.f90
+++ b/src/NavierStokes.f90
@@ -255,7 +255,7 @@ contains
 !c     navier-stokes equations for constant properties
 !c***********************************************************************
       INTEGER (dp) :: i, j, k, g, n ,nx_var,ny_var,nz_var, n1, nn, i11, j11, k11, &
-           index_ip1, index_im1, index_jp1, index_jm1, index_kp1, index_km1
+           index_ip1, index_im1, index_jp1, index_jm1, index_kp1, index_km1, block_idx_ts, idx
          REAL (dp) :: dpdx,dpdy,dpdz,u1a,u22,u3,u4,u5,u6,u7,u8,u9,u10,u11,u12,u13,u14, &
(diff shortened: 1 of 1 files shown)


tomneep:
No file writing, 5 iterations, commit de2acf77a08a12bb84c2d652b299050bba9a2bfd (dev branch)
```
Time (%)  Total Time (ns)  Instances     Avg (ns)         Med (ns)        Min (ns)       Max (ns)      StdDev (ns)                         Name
 --------  ---------------  ---------  ---------------  ---------------  -------------  -------------  -------------  ------------------------------------------------
12.9    1,823,545,840         15    121,569,722.7     62,869,786.0     52,555,418    540,481,518  123,152,976.0  biocfd_navier_stokes_nsmomentum2order_298_gpu
```

With this PR
```
12.0    1,680,507,491         15    112,033,832.7     61,229,114.0     47,545,773    522,768,091  120,390,936.5  biocfd_navier_stokes_nsmomentum2order_298_gpu
```

I'm not sure I'd trust this as a significant change in speed, but I'm pleased to see there is no detrimental impact.

src/NavierStokes.f90
```diff
@@ -330,434 +330,182 @@ subroutine nsMomentum2order
            block(g)%cell2(i, j, k+1)==2.OR. block(g)%cell2(i, j, k-1)==2) THEN
         !$acc loop seq
              DO nn = 1, block(g)%TSCellCount
+                ! TODO: i11, j11, and k11, aren't used anywhere
```
mcbarton: I'm assuming that i11, j11, and k11 can be removed as part of this PR if they aren't use anywhere?

tomneep: Yeah good point, I've removed them now

tomneep: Linter issues

tomneep: Hopefully resolved now

mcbarton (approved):

PR #125, opened 2025-07-07
Remove unneeded sqrt calculation from retagging

tomneep:
There is no need to perform a `sqrt` inside this part of `selectiveretagging_th`. We have done the same in the `tagging_th` subroutine previously.

Side-by-side diff of kernel stats from running 20 iterations on Baskerville (A100). Old on left, new on right.

```diff
selectiveretagging_th  NVIDIA  devicenum=0                      selectiveretagging_th  NVIDIA  devicenum=0
    time(us): 52,458,322                                      |     time(us): 20,849,843
```

Edit: I also removed `sqrt`s from a similar block of code in `computenormdistance`

### DIFF
--- a/src/Search.f90
+++ b/src/Search.f90
@@ -728,8 +728,7 @@ module biocfd_search
         INTEGER(int64) ::  n, g, m, i, j, k, i1, j1, k1, nn, &
                                nel2Pnt, nel2Cen, sumNodeID
         INTEGER            :: flcnt, sdcnt, ibcnt
-        REAL(dp)      :: n1x, n1y, n1z, n2x, n2y, n2z, minDis, minDis1, dis_cen, dis_pnt, &
-                              n2dotn, cent_x, cent_y, cent_z
+        REAL(dp)      :: minDis, minDis1, dis_cen, dis_pnt, n2dotn
 
        DO g=blk_start,nblocks
         if( block(g)%blk_mv_tag ==0)then
--- a/src/Search.f90
+++ b/src/Search.f90
@@ -748,6 +748,7 @@ module biocfd_search
 
                !$acc loop seq
                DO m = 1, block(g)%ibElems
+                  ! No need to take sqrt because we just use for distance comparison
                   dis_cen  = (block(g)%xp(i)-block(g)%xcent(m))**2 &
                            + (block(g)%yp(j)-block(g)%ycent(m))**2 &
                            + (block(g)%zp(k)-block(g)%zcent(m))**2
--- a/src/Search.f90
+++ b/src/Search.f90
@@ -748,11 +748,11 @@ module biocfd_search
 
                !$acc loop seq
                DO m = 1, block(g)%ibElems
-                  dis_cen  = (block(g)%yp(j)-block(g)%ycent(m))**2 &
-                           + (block(g)%xp(i)-block(g)%xcent(m))**2  &
+                  dis_cen  = (block(g)%xp(i)-block(g)%xcent(m))**2 &
+                           + (block(g)%yp(j)-block(g)%ycent(m))**2 &
                            + (block(g)%zp(k)-block(g)%zcent(m))**2
-                  dis_pnt  = (block(g)%y1(j)-block(g)%ycent(m))**2 &
-                           + (block(g)%x1(i)-block(g)%xcent(m))**2  &
+                  dis_pnt  = (block(g)%x1(i)-block(g)%xcent(m))**2 &
+                           + (block(g)%y1(j)-block(g)%ycent(m))**2 &
                            + (block(g)%z1(k)-block(g)%zcent(m))**2
                   IF (dis_cen<minDis) THEN
                      minDis    = dis_cen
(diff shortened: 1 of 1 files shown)


src/Search.f90
```diff
@@ -747,21 +746,14 @@ SUBROUTINE selectiveRetagging_th
...
-               n2z = block(g)%z1(k)
-
                !$acc loop seq
                DO m = 1, block(g)%ibElems
-               cent_x = block(g)%xcent(m)
-               cent_y = block(g)%ycent(m)
-               cent_z = block(g)%zcent(m)
-                  dis_cen  = dsqrt( (n1y-cent_y)**2 + (n1x-cent_x)**2  + (n1z-cent_z)**2)
-                  dis_pnt  = dsqrt( (n2y-cent_y)**2 + (n2x-cent_x)**2  + (n2z-cent_z)**2)
+                  dis_cen  = (block(g)%yp(j)-block(g)%ycent(m))**2 &
```
tomneep: These changes mirror those made in `tagging_th` previously

tomneep: The crucial part is removing the `sqrt` calls, which aren't needed

src/Search.f90
```diff
@@ -978,13 +970,14 @@ SUBROUTINE computeNormDistance
...
-              dis5  = dsqrt( (n1y-cent_y)**2 + (n1x-cent_x)**2 + (n2z-cent_z)**2 )
-              dis6  = dsqrt( (n1y-cent_y)**2 + (n1x-cent_x)**2 + (n3z-cent_z)**2 )
+        ! No need to take sqrt because we just use for distance comparison
+              dis   =  (n1y-cent_y)**2 + (n1x-cent_x)**2 + (n1z-cent_z)**2
+              dis1  =  (n1y-cent_y)**2 + (n2x-cent_x)**2 + (n1z-cent_z)**2
+              dis2  =  (n1y-cent_y)**2 + (n3x-cent_x)**2 + (n1z-cent_z)**2
+              dis3  =  (n2y-cent_y)**2 + (n1x-cent_x)**2 + (n1z-cent_z)**2
+              dis4  =  (n3y-cent_y)**2 + (n1x-cent_x)**2 + (n1z-cent_z)**2
+              dis5  =  (n1y-cent_y)**2 + (n1x-cent_x)**2 + (n2z-cent_z)**2
+              dis6  =  (n1y-cent_y)**2 + (n1x-cent_x)**2 + (n3z-cent_z)**2
```
warrickball: Very minor but the enumeration is a bit easier to understand if we order the sum `xyz` instead of `yxz` (unless there's a good reason for the latter):
```suggestion
              dis   =  (n1x-cent_x)**2 + (n1y-cent_y)**2 + (n1z-cent_z)**2
              dis1  =  (n2x-cent_x)**2 + (n1y-cent_y)**2 + (n1z-cent_z)**2
              dis2  =  (n3x-cent_x)**2 + (n1y-cent_y)**2 + (n1z-cent_z)**2
              dis3  =  (n1x-cent_x)**2 + (n2y-cent_y)**2 + (n1z-cent_z)**2
              dis4  =  (n1x-cent_x)**2 + (n3y-cent_y)**2 + (n1z-cent_z)**2
              dis5  =  (n1x-cent_x)**2 + (n1y-cent_y)**2 + (n2z-cent_z)**2
              dis6  =  (n1x-cent_x)**2 + (n1y-cent_y)**2 + (n3z-cent_z)**2
```

src/Search.f90
```diff
@@ -747,21 +746,14 @@ SUBROUTINE selectiveRetagging_th
...
-               cent_y = block(g)%ycent(m)
-               cent_z = block(g)%zcent(m)
-                  dis_cen  = dsqrt( (n1y-cent_y)**2 + (n1x-cent_x)**2  + (n1z-cent_z)**2)
-                  dis_pnt  = dsqrt( (n2y-cent_y)**2 + (n2x-cent_x)**2  + (n2z-cent_z)**2)
+                  dis_cen  = (block(g)%yp(j)-block(g)%ycent(m))**2 &
+                           + (block(g)%xp(i)-block(g)%xcent(m))**2  &
+                           + (block(g)%zp(k)-block(g)%zcent(m))**2
+                  dis_pnt  = (block(g)%y1(j)-block(g)%ycent(m))**2 &
+                           + (block(g)%x1(i)-block(g)%xcent(m))**2  &
+                           + (block(g)%z1(k)-block(g)%zcent(m))**2
```
warrickball: This is very much nitpicking but I did a double-take when I saw the `yxz` order of the sum, rather than `xyz`.
```suggestion
                  dis_cen  = (block(g)%xp(i)-block(g)%xcent(m))**2 &
                           + (block(g)%yp(j)-block(g)%ycent(m))**2 &
                           + (block(g)%zp(k)-block(g)%zcent(m))**2
                  dis_pnt  = (block(g)%x1(i)-block(g)%xcent(m))**2 &
                           + (block(g)%y1(j)-block(g)%ycent(m))**2 &
                           + (block(g)%z1(k)-block(g)%zcent(m))**2
```

tomneep:
Agree about the xyz ordering. I'd left it the same as it was but I think it makes sense to change it here now. Done!

warrickball (approved):

tomneep:
Thanks @warrickball. @mcbarton can you merge this when you have a chance and then I'll retarget #126.

src/Search.f90
```diff
@@ -747,21 +746,14 @@ SUBROUTINE selectiveRetagging_th
...
-               n2z = block(g)%z1(k)
-
                !$acc loop seq
                DO m = 1, block(g)%ibElems
-               cent_x = block(g)%xcent(m)
-               cent_y = block(g)%ycent(m)
-               cent_z = block(g)%zcent(m)
-                  dis_cen  = dsqrt( (n1y-cent_y)**2 + (n1x-cent_x)**2  + (n1z-cent_z)**2)
-                  dis_pnt  = dsqrt( (n2y-cent_y)**2 + (n2x-cent_x)**2  + (n2z-cent_z)**2)
+                  dis_cen  = (block(g)%xp(i)-block(g)%xcent(m))**2 &
```
mcbarton, line 752: @tomneep Given we are no longer taking the square root, then this is no longer a distance, so the variable name may now be misleading, and make people looking at the code think its a mistake. Could we rename the variables?

tomneep: I've added a comment which I think is enough for now. We've already done this once without changing the variable name.

This code pattern appears in many of the subroutines in this module so it would suggest we'd need to change it everywhere.

mcbarton (approved):

PR #126, opened 2025-07-10
Optimise retagging loops for GPU

tomneep:
Merge after #125 (make sure that we are targeting the right branch then!).

This allows us to remove a serial triple-nested loop.

- [x] Demonstrate performance (the impact is large, but will add some proper numbers once #125 is merged)

I'd like to do some more output diff checks but this is the performance change, selective-retagging-speed (left) v this branch (right)

```
  selectiveretagging_th  NVIDIA  devicenum=0			  selectiveretagging_th  NVIDIA  devicenum=0
    time(us): 20,733,667				      |	    time(us): 1,926,856
```

### DIFF
--- a/src/Search.f90
+++ b/src/Search.f90
@@ -728,19 +728,30 @@ module biocfd_search
                                nel2Pnt, nel2Cen, sumNodeID
         INTEGER            :: flcnt, sdcnt, ibcnt
         REAL(dp)      :: minDis, minDis1, dis_cen, dis_pnt, n2dotn
+        integer :: iprime, jprime, kprime
 
        DO g=blk_start,nblocks
         if( block(g)%blk_mv_tag ==0)then
-!$acc parallel loop gang vector default(present)
+
+       ! Set the intercepted indicies cell value to 0, we do this in a
+       ! seperate loop so that we can nicely GPU-ise the computation
+       !$acc parallel loop gang vector default(present) private(i1, j1, k1)
+       DO nn = 1, block(g)%ibCellCount
+         i1 = block(g)%interceptedIndexPtr(nn, 1)
+         j1 = block(g)%interceptedIndexPtr(nn, 2)
+         k1 = block(g)%interceptedIndexPtr(nn, 3)
+         block(g)%cell(i1,j1,k1) = 0
+       END DO
+
+!$acc parallel loop gang vector default(present) collapse(4) private(i1, j1, k1, iprime, jprime, kprime)
         DO nn = 1, block(g)%ibCellCount
-        i1 = block(g)%interceptedIndexPtr(nn, 1)
-        j1 = block(g)%interceptedIndexPtr(nn, 2)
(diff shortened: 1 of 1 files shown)


tomneep:
This is ready for review now that #125 has been merged. Just to summarise this combined with #125 reduces the time spent in `selectiveretagging_th` from > 52 seconds to < 2 seconds (for 20 iterations).

src/Search.f90
```diff
@@ -728,19 +728,31 @@ SUBROUTINE selectiveRetagging_th
...
+       !$acc parallel loop default(present) private(i1, j1, k1)
+       DO nn = 1, block(g)%ibCellCount
+         i1 = block(g)%interceptedIndexPtr(nn, 1)
+         j1 = block(g)%interceptedIndexPtr(nn, 2)
+         k1 = block(g)%interceptedIndexPtr(nn, 3)
+         block(g)%cell(i1,j1,k1) = 0
+        END DO
+        !$acc end parallel loop
+
+!$acc parallel loop default(present) collapse(4) private(i, j, k)
```
warrickball: I have a feeling I'm missing something simple here but if we're collapsing four loops, why isn't `nn` also private?

tomneep: It is a good question. The answer is that loop variables in this case (`nn`, `kprime`, `jprime`, and `iprime`) are always private. We could be more explicit about some of the other variables e.g. `minDis`. I'll check what OpenACC is doing.

tomneep: So it is doing "the right thing" (really a bunch of the variables can be `private` rather than `firstprivate`) with the variables however I'm going to make it explicit for the variables inside the loop, particularly because of the Accelerator restriction warnings. I think I'll leave the loop variables as implicit, unless there are strong objections to that?

```
    747, Generating implicit firstprivate(g)
         Generating NVIDIA GPU code
        748, !$acc loop gang, vector(128) collapse(4) ! blockidx%x threadidx%x
        749,   ! blockidx%x threadidx%x collapsed
        750,   ! blockidx%x threadidx%x collapsed
        751,   ! blockidx%x threadidx%x collapsed
        761, !$acc loop seq
    747, Generating default present(block(:))
    751, Generating implicit firstprivate(mindis,n2dotn,nel2cen,mindis1,nel2pnt)
    761, Generating implicit firstprivate(dis_pnt,dis_cen)
    771, Accelerator restriction: induction variable live-out from loop: nel2cen
    775, Accelerator restriction: induction variable live-out from loop: nel2pnt
```

tomneep: I've added them now

warrickball: I'm happy as long as whatever we're doing is reasonably close to idiomatic / common practice. Should we then leave, `i`, `j` and `k` implicit too?

tomneep: No because they aren't the loop variables, `[ijk]prime` are 🙂

warrickball: :facepalm:

mcbarton (approved):

Issue #116, opened 2025-05-27
Double check whether this should be `w`


tomneep:
https://github.com/BioFSILab/Bio-CFD/blob/05eea4c98ed1a402a3a24fec09f7704fbf1748c6/src/CoarseInterp.f90#L429-L432

I'm unsure why this is `w` and whether it should be or not. It looks to me like this could be a bug.

warrickball:
FWIW, just going by the code without understanding much of it, it looks to me like it should be `wt` instead of `w`. There's [an earlier block](https://github.com/BioFSILab/Bio-CFD/blob/05eea4c98ed1a402a3a24fec09f7704fbf1748c6/src/CoarseInterp.f90#L220-L252) that looks like it's doing some kind of interpolation to assign `w` between blocks. This block then looks like it does the same thing for `wt`.

tomneep:
Yep, I'm 99.99% sure it should be `wt` (perhaps I should have made that more obvious in the initial report!)

tomneep:
In case things start to diverge, the location of this issue in the main branch is here: 

https://github.com/BioFSILab/Bio-CFD/blob/ba49203ad1f61a2c9ca278e20ddd47d2e0db85e6/src/subCoarseInterp.f90#L854-L857

SiddharthDS:
Yeah. It should be wt instead of w.

tomneep:
Thanks @SiddharthDS! We will fix that our end

PR #128, opened 2025-07-21
Coarse interpolation refactor

tomneep:
Refactors the CoarseInterpolation module to use the functions from the Interpolation module. Makes it clearer what calculation is actually being done (trilinear interpolation).

### DIFF
--- a/src/CoarseInterp.f90
+++ b/src/CoarseInterp.f90
@@ -2,6 +2,7 @@ module biocfd_coarse_update
   use, intrinsic :: iso_fortran_env, only: dp => real64
   ! allow(use-all) - TODO: Aim to fix this in the future
   use global
+  use biocfd_interpolation, only: bilinear_interpolation, linear_interpolation
   implicit none
   private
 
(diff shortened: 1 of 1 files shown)


src/CoarseInterp.f90
```diff
@@ -478,49 +288,86 @@ subroutine coarseUpdate_pc
...
+        integer :: ii, jj, kk
+        ! Indices used to determine the variable indices after offsets
+        integer :: vi, vj, vk
+
+        ii = i
+        jj = j
+        kk = k
+        vi = i
+        vj = j
+        vk = k
```
tomneep, line 347: I don't particularly like how I've named these variables, any suggestions?

src/CoarseInterp.f90
```diff
@@ -439,19 +246,22 @@ subroutine coarseUpdate_newv
                      ((bl_interp_ans2 - bl_interp_ans1)/(bl_intp_z2 - bl_intp_z1)))
                 block(a_blk_no)%wt(tar_x,tar_y,tar_z-1)=bl_interp_ans
 
+                ! The following block can replace the above once we have an answer on GH issue #116
+                ! block(a_blk_no)%wt(tar_x, tar_y, tar_z-1) = trilinear_interpolation( &
+                !      block(a_blk_no)%xw(tar_x), block(a_blk_no)%yw(tar_y), block(a_blk_no)%zw(tar_z), &
+                !      loc_x, loc_y, loc_z, &
+                !      block(b_blk_no)%xw, block(b_blk_no)%yw, block(b_blk_no)%zw, &
+                !      3, block(b_blk_no)%wt)
```
tomneep, line 254: We need a resolution to #116 for this. If, as I expect, there is a bug, then we can just uncomment this and delete the old calculation

src/CoarseInterp.f90
```diff
@@ -478,49 +288,86 @@ subroutine coarseUpdate_pc
...
+
+        !> The target position of the trilinear interpolation
+        real(dp), intent(in) :: x, y, z
+        !> Indices used to determine the grid locations
+        integer, intent(in) :: i, j, k
+        !> The grids
+        real(dp), intent(in), dimension(:) :: xgrid, ygrid, zgrid
+        !> The offset value changes depending on whether we are computing p, u, v, or w
+        !> p=0, u=1, v=2, w=3
+        integer, intent(in) :: offset
```
tomneep, line 328: This is a bit of a magic number at the moment with p=0, u=1, v=2, and w=3. Is there a better option? Maybe an `enum` (if it works on the GPU), or some private parameters in this module?

tomneep (commented):
I've checked that this isn't changing the stdout. There is probably a few more format changes that can be done, but I wanted to get the bigger parts reviewed before that. Maybe @warrickball can take a look and give some fortran tips?

There are two reasons for looking at this module now. The first is just general health check and refactoring to remove repetition and clear up the intent, but the second is that quite a lot of time per iteration is spent in these interpolations (and their fine block equivalents). 

I can't see any obvious reason why we wouldn't be able to do run these calculations on the GPU?

src/CoarseInterp.f90
warrickball: I have seen code where parameter integers are defined, like
```f90
integer, parameter :: P_OFFSET = 0
integer, parameter :: U_OFFSET = 1
integer, parameter :: V_OFFSET = 2
integer, parameter :: W_OFFSET = 3
```
which is fine with a small scope but can be annoying for all sorts of reasons. (I stumbled upon [this quite deep dive](https://matthiasnoback.nl/2025/06/fortran-enumeration-part-1/).) It seems the closest thing to `enum`s in Fortran are in the C-interoperability features but they don't actually provide many of the desirable features of an enum anyway. So if the scope is only this procedure or module anyway, I think this is a reasonable option.

src/CoarseInterp.f90
warrickball: I think this is what slowed me down on this review but with 10 days more background thought, I can't think of anything either. :shrug:

warrickball (approved):

tomneep:
@mcbarton, since @warrickball has approved this maybe you can merge? I went against protocol and already merged #136 (after Warrick approved), I don't think it caused any problems.

PR #129, opened 2025-07-21
Hdf5 writer

### LINKED ISSUE #109: HDF5 writer
The output file writing (https://github.com/BioFSILab/Bio-CFD/blob/a3363e38fe4b8302e6c5e411f28165c855737733/src/WriteOutput_corner1.f90#L12) is very slow and produced very large (6.5 GB) files. 

It only does this for the first couple of iterations and then every 200 iterations, but for quick tests with a few iterations this takes up a lot of time.

Suggest we add an HDF5 writer.

mcbarton:
Fixes https://github.com/BioFSILab/Bio-CFD/issues/109

This PR gives the user the option to output of writeOutput1 to a hdf5 file instead of an ascii file. You will need hdf5 installed regardless of whether you are making use of it or not. To build and run the code after this PR you will need fpm version 0.11 and above, since it needs the hdf5 metapackage provided by fpm. fpm makes use of pkg config to find hdf5, so you make need to define a package file yourself, and add the path to it, to your pkg_config_path environment variable. The default is USE_HDF5=0. To output to hdf5 you need to change USE_HDF5=1 in the fpm.toml file.

### DIFF
--- a/app/main.f90
+++ b/app/main.f90
@@ -17,10 +17,11 @@
         use biocfd_last_conditions, only: lastConditions
         use biocfd_coefficient_matrix, only: coefficientMatrix
         use biocfd_navier_stokes, only: non_uni_coeff, nsmomentum2order
+        use biocfd_write_output_corner1, only: body_plot, writeresult
 #if USE_HDF5 == 1
-  use biocfd_write_output_corner1, only: writeOutput1_hdf5, body_plot, writeresult
+        use biocfd_write_output_corner1, only: write_output => write_output_hdf5
 #else
-  use biocfd_write_output_corner1, only: writeOutput1, body_plot, writeresult
+        use biocfd_write_output_corner1, only: write_output => write_output_ascii
 #endif
         use biocfd_forcing, only: pressureForcing1, pressureforcingfield, pressureforcingghost, &
              velocityforcing1, velocityforcingfield, velocityforcingghost
@@ -58,11 +59,7 @@
         CALL non_uni_coeff
         totime = totime + deltat
         st_flag=0
-#if USE_HDF5 == 1
-        CALL writeOutput1_hdf5
-#else
-        CALL writeOutput1
-#endif
+        CALL write_output
         coarse_flcnt_check=0
         print*, 'adam'
         DO
@@ -78,11 +75,7 @@
         CALL poissonSolver
         print *,7
         CALL pressureForcing1
-#if USE_HDF5 == 1
-        CALL writeOutput1_hdf5
-#else
-        CALL writeOutput1
-#endif
+        CALL write_output
         !$acc wait
         CALL writeResult
         !$acc wait
--- a/app/main.f90
+++ b/app/main.f90
@@ -17,7 +17,11 @@
         use biocfd_last_conditions, only: lastConditions
         use biocfd_coefficient_matrix, only: coefficientMatrix
         use biocfd_navier_stokes, only: non_uni_coeff, nsmomentum2order
-        use biocfd_write_output_corner1, only: writeOutput1, body_plot, writeresult
+#if USE_HDF5 == 1
+  use biocfd_write_output_corner1, only: writeOutput1_hdf5, body_plot, writeresult
+#else
+  use biocfd_write_output_corner1, only: writeOutput1, body_plot, writeresult
+#endif
         use biocfd_forcing, only: pressureForcing1, pressureforcingfield, pressureforcingghost, &
              velocityforcing1, velocityforcingfield, velocityforcingghost
         IMPLICIT NONE
@@ -54,7 +58,11 @@
         CALL non_uni_coeff
         totime = totime + deltat
         st_flag=0
+#if USE_HDF5 == 1
+        CALL writeOutput1_hdf5
+#else
         CALL writeOutput1
+#endif
         coarse_flcnt_check=0
         print*, 'adam'
         DO
@@ -70,7 +78,11 @@
         CALL poissonSolver
         print *,7
         CALL pressureForcing1
+#if USE_HDF5 == 1
+        CALL writeOutput1_hdf5
+#else
         CALL writeOutput1
+#endif
         !$acc wait
         CALL writeResult
         !$acc wait
--- a/fpm.toml
+++ b/fpm.toml
@@ -6,7 +6,6 @@ auto-executables = true
 auto-tests = true
 auto-examples = true
 external-modules = ["hdf5"]
-link = ["hdf5", "hdf5_fortran"]
 
 [preprocess]
 cpp.suffixes = ["f90"]
--- a/fpm.toml
+++ b/fpm.toml
@@ -6,6 +6,7 @@ auto-executables = true
 auto-tests = true
 auto-examples = true
 external-modules = ["hdf5"]
+link = ["hdf5", "hdf5_fortran"]
 
 [preprocess]
 cpp.suffixes = ["f90"]
--- a/fpm.toml
+++ b/fpm.toml
@@ -6,12 +6,10 @@ auto-executables = true
 auto-tests = true
 auto-examples = true
 external-modules = ["hdf5"]
-link = ["hdf5", "hdf5_fortran"]
 
 [preprocess]
-cpp.suffixes = ["F90", "f90"]
-cpp.directories = ["src/"]
-cpp.macros = ["USE_HDF5=0"]
+cpp.suffixes = ["f90"]
+cpp.macros = ["USE_HDF5=1"]
 
 [install]
 library = false
--- a/fpm.toml
+++ b/fpm.toml
@@ -1,9 +1,18 @@
 name = "Bio-CFD"
+dependencies.hdf5 = "*"
 version = "0.1.0"
 [build]
 auto-executables = true
 auto-tests = true
 auto-examples = true
+external-modules = ["hdf5"]
+link = ["hdf5", "hdf5_fortran"]
+
+[preprocess]
+cpp.suffixes = ["F90", "f90"]
+cpp.directories = ["src/"]
+cpp.macros = ["USE_HDF5=0"]
+
 [install]
 library = false
 
--- a/fpm.toml
+++ b/fpm.toml
@@ -9,7 +9,7 @@ external-modules = ["hdf5"]
 
 [preprocess]
 cpp.suffixes = ["f90"]
-cpp.macros = ["USE_HDF5=1"]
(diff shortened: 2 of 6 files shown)


fpm.toml
```diff
@@ -1,9 +1,18 @@
...
 version = "0.1.0"
 [build]
 auto-executables = true
 auto-tests = true
 auto-examples = true
+external-modules = ["hdf5"]
+link = ["hdf5", "hdf5_fortran"]
+
+[preprocess]
+cpp.suffixes = ["F90", "f90"]
```
tomneep: I think it might make things clearer if we stick to only pre-processing `F90` files and leave `f90` files untouched.

app/main.f90
```diff
@@ -17,7 +17,11 @@ PROGRAM main
         use biocfd_last_conditions, only: lastConditions
         use biocfd_coefficient_matrix, only: coefficientMatrix
         use biocfd_navier_stokes, only: non_uni_coeff, nsmomentum2order
-        use biocfd_write_output_corner1, only: writeOutput1, body_plot, writeresult
+#if USE_HDF5 == 1
+  use biocfd_write_output_corner1, only: writeOutput1_hdf5, body_plot, writeresult
+#else
+  use biocfd_write_output_corner1, only: writeOutput1, body_plot, writeresult
+#endif
```
tomneep: I suggest here to separate out the `body_plot` and `writeresult` imports
and then for the conditional compilation alias the name of the write e.g.
```f90
use biocfd_write_output_corner1, only: body_plot, writeresult
#if USE_HDF5 == 1
  use biocfd_write_output_corner1, only: write_output => writeOutput1_hdf5 
#else
  use biocfd_write_output_corner1, only: write_output => writeOutput1
#endif
```

Then you won't need any other pre-processor statements in this file

src/WriteOutput_corner1.f90
```diff
@@ -1,14 +1,333 @@
...
+
+      SUBROUTINE writeOutput1_hdf5
+       CHARACTER(len=150)  :: filename1
+       INTEGER  :: k, i, j, g
+       REAL (dp), allocatable :: u1(:,:,:), v1(:,:,:), w1(:,:,:)
+       character (len=11) :: dummy_1
+       character (len=5) ::dummy_2
+
+         IF((mod(ita,200_int64) ==0 .or. ita <= 2 ))then
+         !$acc serial
```
tomneep: Do you know why there are acc directives here? Can we be careful elsewhere to avoid these?

fpm.toml
```diff
@@ -1,9 +1,18 @@
 name = "Bio-CFD"
+dependencies.hdf5 = "*"
 version = "0.1.0"
 [build]
 auto-executables = true
 auto-tests = true
 auto-examples = true
+external-modules = ["hdf5"]
+link = ["hdf5", "hdf5_fortran"]
```
tomneep: [The FPM example](https://fpm.fortran-lang.org/spec/metapackages.html#hdf5) doesn't include this line, are we sure it is needed? I imagine we need the `external-modules` one for the preprocessing.

fpm.toml
```diff
@@ -1,9 +1,18 @@
...
+
+[preprocess]
+cpp.suffixes = ["F90", "f90"]
+cpp.directories = ["src/"]
```
tomneep: This is interesting, because it seems to be doing `app/` too automatically?

src/WriteOutput_corner1.f90
```diff
@@ -1,14 +1,333 @@
...
+#if USE_HDF5 == 1
+  public :: writeOutput1_hdf5, body_plot, writeresult
+#else
   public :: writeOutput1, body_plot, writeresult
+#endif
 
 contains
+
+#if USE_HDF5 == 1
+  subroutine hdf5_write_real(filename,scalar_input,&
```
tomneep: I'd be tempted to put these subroutines into a separate module, so that `WriteOutput_corner1.f90` (maybe needs a better name!) only contains the public subroutines?

src/WriteOutput_corner1.f90
```diff
@@ -1,14 +1,333 @@
...
+
+    ! Close the file.
+    call h5fclose_f(file_id, error)
+
+    ! Close Fortran interface.
+    call h5close_f(error)
+
+  end subroutine hdf5_write_int
+
+      SUBROUTINE writeOutput1_hdf5
```
tomneep: Now might be a good time to give this a nicer name e.g. `write_output_hdf5` (and similarly `write_output_ascii` below)

fpm.toml
tomneep: For the `multi-gpu` branch I was changing files that required preprocessing to have `F90` as the extension, so that we can easily see what is being preprocessed and what isn't

src/WriteOutput_corner1.f90
```diff
@@ -1,14 +1,333 @@
...
+            DO k = 2, block(g)%nz+1
+            DO j = 2, block(g)%ny+1
+            DO i = 2, block(g)%nx+1
+               u1(k,j,i) = 0.5_dp*(block(g)%u(i,j,k)+block(g)%u(i-1,j,k))
+               v1(k,j,i) = 0.5_dp*(block(g)%v(i,j,k)+block(g)%v(i,j-1,k))
+               w1(k,j,i) = 0.5_dp*(block(g)%w(i,j,k)+block(g)%w(i,j,k-1))
+            END DO
+            END DO
+            END DO
+            filename1="output_"//trim(dummy_2)//".h5"
```
tomneep: The ascii output gets written to the `out/` directory. Hopefully we can make this configurable in the future but for now should we write the `h5` files to the same place?

tomneep: Also, did you already discuss with the team about writing one file with all blocks per iteration, rather than what is done for the ascii?

src/WriteOutput_corner1.f90
```diff
@@ -1,14 +1,333 @@
...
+            write(dummy_2,'(I5.5)') ita
+            allocate(u1(2:block(g)%nz+1,2:block(g)%ny+1,2:block(g)%nx+1),&
+                     v1(2:block(g)%nz+1,2:block(g)%ny+1,2:block(g)%nx+1),&
+                     w1(2:block(g)%nz+1,2:block(g)%ny+1,2:block(g)%nx+1))
+            DO k = 2, block(g)%nz+1
+            DO j = 2, block(g)%ny+1
+            DO i = 2, block(g)%nx+1
+               u1(k,j,i) = 0.5_dp*(block(g)%u(i,j,k)+block(g)%u(i-1,j,k))
+               v1(k,j,i) = 0.5_dp*(block(g)%v(i,j,k)+block(g)%v(i,j-1,k))
+               w1(k,j,i) = 0.5_dp*(block(g)%w(i,j,k)+block(g)%w(i,j,k-1))
```
tomneep: Can you add a comment about why this is (z, y, x) i.e. (k, j, i) rather than (x, y, z)? it means that we get different shaped outputs (which maybe what was in the ascii anyway?)

```console
❯ h5ls output_00001.h5/block_00001
cell                     Dataset {256, 256, 374}
cell_n                   Dataset {256, 256, 374}
cell_pr                  Dataset {257, 257, 375}
p                        Dataset {256, 256, 374}
totime                   Dataset {SCALAR}
u1                       Dataset {372, 254, 254}
v1                       Dataset {372, 254, 254}
w1                       Dataset {372, 254, 254}
xp                       Dataset {374}
yp                       Dataset {256}
zonei                    Dataset {SCALAR}
zonej                    Dataset {SCALAR}
zonek                    Dataset {SCALAR}
zp                       Dataset {256}
```

tomneep (commented):
This is already looking good. A few comments/suggestions below. I tried reading one of the output files in Paraview but I have no idea what I'm doing. Have you tested that we can read the files?

fpm.toml
mcbarton: F90 is the default I believe, so I should probably just remove that, and keep f90. Then in your PR you should be able to drop this line altogether since .F90 is preprocessed by default.

app/main.f90
mcbarton: Good idea. I'll implement this.

src/WriteOutput_corner1.f90
mcbarton: I think it was a case of me being cautious, since I am not using any of hdf5 parallel features, so unsure what would happen here if I try to write to the file in parallel.

fpm.toml
mcbarton: I'll remove it, and see if everything still runs fine locally.

fpm.toml
mcbarton: This line is a mistake on my part. I will remove.

src/WriteOutput_corner1.f90
mcbarton: I'll move them to a hdf5_io module.

src/WriteOutput_corner1.f90
mcbarton: Good catch. Will implement.

src/WriteOutput_corner1.f90
mcbarton: I can move it to write in the out directory. I can make this configurable in my namelists PR. I think we discussed with the team about putting everything from one iteration in a single hdf5 file, but will check in our next meeting.

src/WriteOutput_corner1.f90
mcbarton: This may be a mistake on my part. I will investigate.

tomneep: Did this not need changing in the end?

src/WriteOutput_corner1.f90
```diff
@@ -1,15 +1,81 @@
 module biocfd_write_output_corner1
-  use, intrinsic :: iso_fortran_env, only: dp => real64, int64
+  use, intrinsic :: iso_fortran_env, only: dp => real64, int8,int16,int32,int64
```
tomneep: Is `int16` used anywhere?

mcbarton: int64 and real64 are the only ones being used. Removed unused ones.

src/WriteOutput_corner1.f90
mcbarton: This should be fixed now with the latest commit.

tomneep (approved):
I think this looks good to go in now, so I'll approve and merge. We can try and think of a good way to allow any precision for writing integer arrays because we'll probably want to use more sensible types at some point.

PR #130, opened 2025-07-23
Use stepped looping in fineupdate_bd_mv

tomneep:
Follow on from #121 working in the spirit of #33. Adds strided loops like has been done for the other subroutines previously.

Again, the diff is messed up so I recommend also looking at the file


tomneep:
Hi @warrickball, since you had a look at the predecessor to this (#121), would you mind having a look at this one too? It might require flicking back and forth between files because the diff is a bit messed up. I think that once this is merged we can write a function to do most of the work (see #128 for similar).

Once that is done it should be straightforward to get these subroutines running on the GPU, provided we can satisfy ourselves they can be!

warrickball (approved):
LGTM! I did flick back and forth between the branches by hand. Not perfect or ideal but between that and the diff (which was about 2/3 correctly rendered) I hopefully didn't let anything through.

tomneep:
Thanks @warrickball. @mcbarton can you merge when you have time (no rush)?

PR #131, opened 2025-07-28
Multi-GPU (single node) PcorVcor

tomneep:
Opening this up now mostly to keep track of it and make sure that the CI runs ok.
In the end I don't think the changes are too intrusive. 

At the moment the only multi-GPU-ized code is in PcorVcor, where we run with one OMP thread for each block, and the extra overhead of copying data between devices isn't really worth it at the moment. However, if we had lots of butterflies I'm pretty sure we'd get some benefit eventually!


tomneep:
PR not picking up new commits? Commenting to see if it spurs it into action...

warrickball:
As an aside, you can make genuinely completely empty commits with `git commit --allow-empty`. I worked on a project where we occasionally had to prod an external CI service with these.

mcbarton:
@tomneep can you fix the conflict for this PR? I will then review.

src/PcorVcor.F90
```diff
@@ -3,6 +3,12 @@ module biocfd_pcor_vcor
   use global, only : block, deltat, epsi, omega, omega1, omega2, omega3, omega4, pcitamax, &
        amgxita, couptime, dfinish, dstart, ita, mstime, nblocks, solvertime, totaltime, &
        totime
+#ifdef _OPENMP
+  use omp_lib, only: omp_get_max_threads, omp_get_thread_num
```
mcbarton, line 7: Why do we not need 'omp_lib' as an external module?

src/PcorVcor.F90
```diff
@@ -58,40 +87,58 @@ SUBROUTINE poissonSolver
...
-        DO g=2,nblocks
-         CALL cpu_time(dStart)
-         amgxita=0
-         CALL REDBLACKSOR_linear(g)
-        CALL cpu_time(dfinish)
-         msTime = msTime + dfinish-dstart
+           CALL computeDiv(g)    !divergence vector
+           ! Do not compute Red/Black here for block 1
+           if (g == 1) cycle
+           CALL REDBLACKSOR_linear(g)
```
mcbarton:
```suggestion
           if (g /= 1)  CALL REDBLACKSOR_linear(g)

```

src/PcorVcor.F90
tomneep: I don't know, but the CI and my tests all seem to work. Strange.

mcbarton (approved):

mcbarton:
@tomneep Feel free to merge once the ci finishes

PR #135, opened 2025-07-29
Use our existing function for PressureForcingGhost and PressureForcingField

tomneep:
This just uses an existing function that we added in #112 to perform the pressure operations in the forcing module. The function requires a few changes to do the velocity components u, v, and w, but once this is in we can look at that. 

Tested on Baskerville and no change in outputs beyond what we've seen run-to-run before.


tomneep:
@mcbarton I think I've resolved my worries about this so it can be reviewed now (providing the CI all passes)

mcbarton (approved):

Issue #107, opened 2025-05-07
Some of the output changes each run


tomneep:
I see that the output changes slightly. 

The particular line seems to be this one:

https://github.com/BioFSILab/Bio-CFD/blob/a3363e38fe4b8302e6c5e411f28165c855737733/src/PcorVcor.f90#L140

The values that change are `max_derr2` and `max_derrStdSt` e.g. 

```
20       315   0.953223E-05   0.112543E+04   0.185642E+00
```
and 
```
20       315   0.955987E-05   0.112449E+04   0.185642E+00
```

I'm wondering whether this could result from this accelerated section of code

https://github.com/BioFSILab/Bio-CFD/blob/fab84fe20c877a2760b11c1c8622a091426ef37f/src/PcorVcor.f90#L286-L305

as `pc` is being modified in the loop, but also used in the calculation...

tomneep:
Note that removing the acceleration of this block has a big effect on performance

tomneep:
Fixing the parallelisation of that block doesn't seem to fix the changing output, will investigate further

PR #136, opened 2025-07-30
Fine interp bound axes loops

tomneep:
This is the next step in the quest of #33. 

In the subroutines in this module, the code is looping over the six faces of fine (butterfly) cuboids. Before this PR, this was being done two faces at a time (after previous PRs) i.e. in one block of code it would do the "low x" face and the "high x face", then in another code block the "low y" face and "high y" face, and then in a final code block the "low z" face and "high z" face. For each variable, the insides of these three code blocks are identical.

This PR just adds a loop over axes such that all six faces can be addressed in one block of code.

The key bit of code is the pattern like this

```f90
! Loop through the x (1), y (2), and z (3) axes
DO axis=1, 3

  ! Set all steps to 1
  steps = 1

  ! If intfr%counter[xyz]p was a single variable this would be nicer
  if (axis == 1) then
    steps(1) = intfr(g)%counterxp-1
  else if (axis == 2) then
    steps(2) = intfr(g)%counteryp-1
  else if (axis == 3) then
    steps(3) = intfr(g)%counterzp-1
  end if

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!ppppppp!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
        DO k=1, intfr(g)%counterzp, steps(3)
        DO j=1, intfr(g)%counteryp, steps(2)
        DO i=1, intfr(g)%counterxp, steps(1)
```
and repeat for each variable we need to consider.

Once this is in the next steps are
+ Turn the inside of the triple loop into a function (similar to what is being done in #128)
+ Add OpenACC directives to make these loops run on the GPU

Following that there is certainly additional refactoring that could be done, such as replacing `counter[xyz]p` with an array of length 3 `counter_p` but this would touch several other files in the codebase.

### DIFF
--- a/src/FineInterp_bound.f90
+++ b/src/FineInterp_bound.f90
@@ -29,85 +29,32 @@ SUBROUTINE fineUpdate_bd
         INTEGER(int64) :: i,j,k, varx1,varx2, vary1, vary2, tar_x, tar_y, loc_x, &
              loc_y,g, a_blk_no, b_blk_no
         INTEGER(int64) :: varz1,varz2, tar_z, loc_z
-
-
-
-
+        !> axis and steps control how the looping is performed over the x, y, and z axes
+        integer :: axis, steps(3)
 
         DO g=1,intflines
            a_blk_no=intfr(g)%a_blk
            b_blk_no=intfr(g)%b_blk
 
+        ! Loop through the x (1), y (2), and z (3) axes
+        DO axis=1, 3
+
+          ! Set all steps to 1
+          steps = 1
+
+          ! If intfr%counter[xyz]p was a single variable this would be nicer
+          if (axis == 1) then
+               steps(1) = intfr(g)%counterxp-1
+          else if (axis == 2) then
+               steps(2) = intfr(g)%counteryp-1
+          else if (axis == 3) then
+               steps(3) = intfr(g)%counterzp-1
+          end if
+
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!ppppppp!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-        DO k=1,intfr(g)%counterzp
-        DO j=1,intfr(g)%counteryp
-        DO i=1,intfr(g)%counterxp, intfr(g)%counterxp-1
-
-
-        varx1=intfr(g)%px_interface_det(2,i)
-        varx2=intfr(g)%px_interface_det(3,i)
-        vary1=intfr(g)%py_interface_det(2,j)
-        vary2=intfr(g)%py_interface_det(3,j)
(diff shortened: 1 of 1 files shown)


src/FineInterp_bound.f90
```diff
@@ -387,74 +220,28 @@ SUBROUTINE fineUpdate_newv_bd
...
-                bl_intp_deno= (bl_intp_x2-bl_intp_x1) * (bl_intp_y2-bl_intp_y1)
-                bl_intp_xtx= (bl_intp_x2 -bl_intp_valx)
-                bl_intp_xxo=(bl_intp_valx-bl_intp_x1)
-                bl_intp_yty=(bl_intp_y2-bl_intp_valy)
-                bl_intp_yyo=(bl_intp_valy-bl_intp_y1)
-                bl_intp_first_term=(bl_intp_f1*bl_intp_xtx + bl_intp_f2*bl_intp_xxo)*bl_intp_yty
-                bl_intp_second_term=(bl_intp_f4*bl_intp_xtx + bl_intp_f3*bl_intp_xxo)*bl_intp_yyo
-                bl_intp_num= bl_intp_first_term + bl_intp_second_term
+        ! Loop through the x (1), y (2), and z (3) axes
+        DO axis=1, 3
```
tomneep, line 229: We could have a single `axis` loop that goes through u, v, w. At the moment I've kept it separate so that the order of operations is the same as previously

tomneep:
Checked on Baskerville for 200 iterations and no change in stdout (beyond what we've observed previously (#107), although I'm not 100% sure whether the stdout includes anything that would be affected by this.

I'll mark this as ready. @warrickball maybe one for you again since you looked at the previous PR for this module and @mcbarton is very busy with #129? Happy to talk about it in person if it helps.

src/FineInterp_bound.f90
warrickball: Ah, I took a moment to wrap my head around the code here. You mean go from the proposal in this PR:
```f90
do axis=1,3
    ! interpolate p(axis)
end do
do axis=1,3
    ! interpolate u(axis)
end do
do axis=1,3
    ! interpolate v(axis)
end do
do axis=1,3
    ! interpolate w(axis)
end do
```
to
```f90
do axis=1,3
    ! interpolate p(axis)
    ! interpolate u(axis)
    ! interpolate v(axis)
    ! interpolate w(axis)
end do
```
? Agreed that's sensible for a subsequent PR and it's worth checking output. I'd be surprised/worried if it *did* change output because it looks like the result in each loop goes into distinct variables in the block (`pc` and `pco` for `p` then `{u,v,w}t` for `{u,v,w}`, resp.).

Current change looks good though.

warrickball (approved):

PR #140, opened 2025-08-06
Move Block type to its own module

mcbarton:
This is the first step in stopping BioCFD being dependent on global variables. This PR moves the block type to its own module. The next step will be change main, so that only parts of block are used as input to subroutines, rather than just relying on block being a global variable.

### DIFF
--- a/app/main.f90
+++ b/app/main.f90
@@ -25,7 +25,6 @@
 #endif
         use biocfd_forcing, only: pressureForcing1, pressureforcingfield, pressureforcingghost, &
              velocityforcing1, velocityforcingfield, velocityforcingghost
-        use biocfd_blocks, only : block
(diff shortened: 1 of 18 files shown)


mcbarton:
As I removed alpha3, beta3 and gamma3 while moving block, this PR is dependent on https://github.com/BioFSILab/Bio-CFD/pull/139 to pass the ci.

tomneep (changes requested):
I don't think we should do everything you are doing here. I think it is less intrusive to just move the type definition of `Blocks` to a new module as you have done, but then leave the `type(Blocks),allocatable :: block(:)` in `modGlobal` for now.

Then you only need to touch two files, the new Block file and `modGlobal`.

tomneep (approved):

PR #143, opened 2025-08-07
Comment out HDF5 dependency

tomneep:
So that we can build/run without having HDF5 available. I've tested this on Baskerville and it works.

### DIFF
--- a/fpm.toml
+++ b/fpm.toml
@@ -1,5 +1,7 @@
 name = "Bio-CFD"
-dependencies.hdf5 = "*"
+# To use HDF5 uncomment the dependencies.hdf5 line below and set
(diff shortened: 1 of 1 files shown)


mcbarton:
Maybe add a comment to the readme about needing to uncomment this line if you want to use hdf5 (and switching use_hdf5 to be equal to 1). Although I can make this a separate PR if you'd like.

tomneep:
I'd save that for a later date. I think we can document HDF5 when it is usable, which I think requires a way of running it on Baskerville/BlueBEAR, and a script which allows the HDF5 files to be read with Paraview.

mcbarton (approved):

PR #144, opened 2025-08-08
Use a function in FineInterp_bound

### LINKED ISSUE #33: Repetition in `subFineInterp_bound.f90`

tomneep:
This is the next step in reducing the repetition in `FineInterp_bound`. I think we could perhaps even argue that this would maybe close #33, as removing the remaining repetition is a bit more involved, requiring us to change data structures.

## What's in this PR
- A new pure function `trilinear_interpolation`, that is very similar to what I did in #128 but ever so slightly different (a future task could be unifying them)
- Replace the "by-hand" interpolation in every subroutine with a call to the new function
- For good measure, I've added the OpenACC directives so that we can run this all on the GPU

I went through and tested that we don't change any of the values we set to e.g. `block(g)%p`, there are occasional small differences I put down to machine precision, but the majority of the time the results were exact.

As an aside
```console
❯ git show main:src/subFineInterp_bound.f90 | wc -l
    6989
❯ cat src/FineInterp_bound.f90 | wc -l
     581
```

### DIFF
--- a/src/FineInterp_bound.f90
+++ b/src/FineInterp_bound.f90
@@ -48,6 +48,9 @@ SUBROUTINE fineUpdate_bd
           end if
 
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!ppppppp!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+        !$acc parallel loop collapse(3) private(varx1, varx2, vary1, vary2, varz1, varz2) &
+        !$acc private(loc_x, loc_y, loc_z, bl_interp_ans) &
+        !$acc firstprivate(a_blk_no, b_blk_no)
         DO k=1, intfr(g)%counterzp, steps(3)
         DO j=1, intfr(g)%counteryp, steps(2)
         DO i=1, intfr(g)%counterxp, steps(1)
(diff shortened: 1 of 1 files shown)


tomneep:
Assigning @warrickball as reviewer since he has looked at the PRs leading up to this one, but there is no real rush!

warrickball (approved):
I possibly haven't looked as closely as I would with more time but I've looked at the rough idea in the old and new code and it looks good. That is, I can see each copy+pasted block of the old code basically going through the same order of steps to construct the trilinear interpolation, and that your code implements that trilinear interpolation with the correct arguments.

If it isn't too onerous, I think this might be a good time to add a test for your new trilinear interpolation. It's a very tidy function so I hope it isn't too tricky. But you're welcome to do it in a fresh PR because adding stuff here will presumably reset the approval while I'm away.

tomneep:
Agreed now is a good time to add a test. I'll merge this now and follow up with a test next week. A bit of thought needed on how we do the testing because this isn't a public function, so won't currently be accessible from another module.

PR #145, opened 2025-08-08
Replace reading inputdata file with a namelist file for input_data

mcbarton:
This PR changes the input to be based on namelists.

### DIFF
--- a/.github/workflows/setup-fortran.yml
+++ b/.github/workflows/setup-fortran.yml
@@ -39,6 +39,8 @@ jobs:
           # Compile with OpenACC since there are some directives
           - {compiler: gcc, version: 14, FPM_FFLAGS: "-fopenacc", tag: "-openacc"}
           - {compiler: nvidia-hpc, version: '25.1', FPM_FFLAGS: "-acc=multicore", tag: "-openacc"}
+          # Cannot currently support lfortran due to it not supporting namelists (issue number 1999 lfortran)
+          # - {compiler: lfortran, version: '0.55.0'}
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/setup-fortran.yml
+++ b/.github/workflows/setup-fortran.yml
@@ -39,7 +39,6 @@ jobs:
           # Compile with OpenACC since there are some directives
           - {compiler: gcc, version: 14, FPM_FFLAGS: "-fopenacc", tag: "-openacc"}
           - {compiler: nvidia-hpc, version: '25.1', FPM_FFLAGS: "-acc=multicore", tag: "-openacc"}
-          - {compiler: lfortran, version: '0.55.0'}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


mcbarton:
Example namelist file

```
&input_data
nblocks=3, 
intflines=2,  
itamax=1, 
epsi=1e-5, 
pcItaMax=35000,
omega1=1.98,
omega2=1.94,
omega3=1.94,
omega4=1.9,
re=500,
rho_f=1.225, 
mu_f=1.804, 
l_c=56.3,
u0=460,
v0=0,
w0=0.0,
surGeoPoints=0,
a0y=15,
phase_angle=90.0,
freq=0.4,
aoa=30,
piv_pt=0.0,
alpha_m=0.0,
theta_m=45.0,
istart=0,
dt_order=1000, 
inor=1,
dxmin=1.25
/
```

src/ReadInput.f90
```diff
@@ -23,14 +23,17 @@ SUBROUTINE readInput
...
-        READ(60,*) nblocks, intflines,  &
+       NAMELIST /input_data/ nblocks, intflines,  &
                    itamax, epsi, pcItaMax,omega1,omega2,omega3,omega4, &
                    re,rho_f, mu_f, l_c, &
                    u0, v0,  w0,  &
                    surGeoPoints,a0y, phase_angle, freq, aoa, piv_pt,alpha_m, theta_m, &
                    istart, dt_order, inor, dxmin
-        CLOSE(60)
+
+  OPEN(UNIT=10, FILE='input_data.nml', FORM='formatted')
```
tomneep: Since this is changing, does it make sense to make this use best practices i.e as recommended [here](https://fortran-lang.org/learn/best_practices/file_io/#file-input-output) use `newunit`, `status`, and `action`?

.github/workflows/setup-fortran.yml
```diff
@@ -39,7 +39,6 @@ jobs:
           # Compile with OpenACC since there are some directives
           - {compiler: gcc, version: 14, FPM_FFLAGS: "-fopenacc", tag: "-openacc"}
           - {compiler: nvidia-hpc, version: '25.1', FPM_FFLAGS: "-acc=multicore", tag: "-openacc"}
-          - {compiler: lfortran, version: '0.55.0'}
```
tomneep, line 42: Maybe we can just comment this out (like we had before #149), with a comment saying why we can't use it (issue 1999 in lfortran repo - being ultra cautious and deliberately not putting the link so we don't flag their repo when we make this public)

tomneep (commented):
This works fine and with the example you give I get the same results as the previous input data file. I think it is probably worth keeping the lfortran line in the actions commented out.

tomneep:
Ah and should probably update https://github.com/BioFSILab/biocfd-test-input (I'd add the `nml` as an additional file for now until we merge dev into main)

tomneep (approved):

PR #146, opened 2025-08-11
Unused block variables

tomneep:
These variables in the `Blocks` type are unused (sometimes set but not read)


tomneep:
There's probably more that can be done here, but this seems like plenty to get started with.

mcbarton (approved):

PR #147, opened 2025-08-11
Unused forcing variables

tomneep:
The variables `ac_x, ac_y, ac_z, at_x, at_y, at_z` are only used as private variables in the Forcing module and nowhere else, so they can be removed from global.


mcbarton:
@tomneep Is there a reason this is still set to draft? If not, can you please fix the merge conflict.

tomneep:
You are right I think this can be reviewed now

mcbarton (approved):

PR #148, opened 2025-08-12
Fix bounds check issues

### LINKED ISSUE #118: Fortran runtime error: Index '0' of dimension 3 of array 'block...%cell2' below lower bound of 1
https://github.com/BioFSILab/Bio-CFD/blob/82629445000e7fafd565045cd9e40f6cbdd3c6f3/src/Search.f90#L662-L664

As raised in #106, there is a loop starting at 0 when it should start at 1. This hasn't been picked up so far because we've been running with optimisations on which I think probably skip bounds checks. Adding the flag `-fcheck=all` when compiling with `gfortran` indeed crashes with an error.

```
At line 662 of file ././src/Search.f90
Fortran runtime error: Index '0' of dimension 3 of array 'block...%cell2' below lower bound of 1
```

Probably a good idea to do a run through with `-fcheck=all` after this is fixed to see if there is anything else that is problematic.

tomneep:
This addresses the out-of-bounds errors and will close #118. 

There are fixed in two places. 

1. In Search, where a loop was starting at `0`. Looking at allocate arrays we can see that this is likely just a typo and the iteration should start at 1.
2. In NavierStokes, where the lack of short-circuiting if statements in fortran means that out-of-bounds access was occurring where it wouldn't if short circuiting was present. This is fixed along with a refactor to reduce a bit of repetition.

### DIFF
--- a/src/NavierStokes.f90
+++ b/src/NavierStokes.f90
@@ -312,7 +312,8 @@ contains
 !$acc           temp_v2dotn,temp_v1dotn, &
 !$acc           temp_w2dotn,temp_w1dotn, temp_pdotn) &
 !$acc default(present)   &
-!$acc firstprivate(nx_var, ny_var, nz_var, rev, deltat, al)
+!$acc firstprivate(nx_var, ny_var, nz_var, rev, deltat, al) &
+!$acc private(do_third_order_upwinding)
      DO n = 1, block(g)%fluidCellCount
 
       i = block(g)%fluidIndexPtr(n, 1)
--- a/src/NavierStokes.f90
+++ b/src/NavierStokes.f90
@@ -1,5 +1,5 @@
 module biocfd_navier_stokes
-  use, intrinsic :: iso_fortran_env, only: dp => real64
+  use, intrinsic :: iso_fortran_env, only: dp => real64, int64
   use global, only : block, al, alpha, deltat, nblocks, re, rev
   implicit none
   private
@@ -287,6 +287,7 @@ contains
                        duwdx,dvwdy,dwwdz,d2wdx2,d2wdy2,d2wdz2,xtt2,residu,ytt2,residv,&
                           ztt2,residw, temp_u1dotn, temp_u2dotn, temp_v1dotn, temp_v2dotn, &
                           temp_w1dotn, temp_w2dotn, temp_pdotn
+         logical :: do_third_order_upwinding
      al = 1.
 
         DO g=1,nblocks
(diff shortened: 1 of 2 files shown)


tomneep:
Marked as draft because it needs checking on Baskerville. I think that probably a `!$acc private(do_third_order_upwinding)`would be nice (even if OpenACC will likely do the right thing).

tomneep:
Tested on Baskerville and works fine. Should maybe check output changes more carefully.

src/NavierStokes.f90
```diff
@@ -953,14 +960,19 @@ subroutine nsMomentum2order
...
-      block(g)%cell2(i,j,k-2)/=2) THEN
+
+       do_third_order_upwinding = .false.
+
+       if(i>2 .and. i<nx_var+1 .and. j>2 .and. j<ny_var .and. k>2 .and. k<nz_var+1) then
+          if (check_adjacent_cell(block(g)%cell, i, j, k, 1)) then
+             if (check_adjacent_cell(block(g)%cell2, i, j, k, 2)) then
+                do_third_order_upwinding = .true.
+             end if
+          end if
```
warrickball, line 971: Though I don't think it's a big improvement, this looks to me like
```suggestion
          do_third_order_upwinding = &
             check_adjacent_cell(block(g)%cell, i, j, k, 1) .and. &
             check_adjacent_cell(block(g)%cell2, i, j, k, 2)
```
I think this is just because I'm a fan of inlining operations on `logical`s. You're welcome to ignore this! (I'm also not 100% sure you don't need more brackets but I did some basic testing that suggests this is fine.)

tomneep: That looks good. It probably isn't worth worrying about but do you know if both statements be evaluated or not. At this point it is safe to but maybe it isn't worth checking `cell2` if `cell` doesn't satisfy the condition?

warrickball: AFAIK whether things are all evaluated and in what order, or if the logic short-circuits, is not specified in the language so depends on the compiler. 

That's also a valid reason to reject my suggestion. If you happen to end up near some kind of edge where the `step+2` call will be out of bounds, then your way of folding the logic is better. I've worked on projects where we specifically went from the equivalent of my suggestion back to your nested version to guarantee that one condition was evaluated first.

tomneep: I think given the improvement is so minor (I do think it looks nicer in general), I'll leave this out for now. If there was well defined short circuiting then I think I'd be more in favour...

tomneep:
Hi @warrickball, I know you've got a busy week, but I think you've already looked at this? If you're happy for me or @mcbarton to merge please let me know!

mcbarton (approved):

warrickball:
Sorry I'm late getting back to this but for the record, LGTM.

PR #149, opened 2025-08-13
Lfortran build

tomneep:
A couple of small fixes to get lfortran up and running. I don't expect the small change in `InterfaceDetail` to cause issues, but probably best to make sure this runs on Baskerville before merging.


tomneep:
Quickly checked on Baskerville. All seems fine as expected. Can be reviewed/merged when convenient.

mcbarton (approved):

PR #152, opened 2025-08-22
Fix bug in coarse interpolation with refactor

### LINKED ISSUE #116: Double check whether this should be `w`

tomneep:
Closes #116. 

Now that we've had confirmation, we can apply the refactor we've used everywhere else to the `wt` calculation.


mcbarton (approved):
Looks good to me

PR #159, opened 2025-09-05
Offload coarse interpolation to the GPU

tomneep:
Unless I'm missing something, there is no reason we can't do the coarse interpolation on the GPU. This adds it. I'll make an update to #91 once this is in.

### DIFF
--- a/src/CoarseInterp.f90
+++ b/src/CoarseInterp.f90
@@ -25,6 +25,7 @@ subroutine coarseUpdate
         st_idz=block(b_blk_no)%cintp
         en_idz=intfr(g)%counterzp-block(b_blk_no)%cintp
 
+        !$acc parallel loop collapse(3) private(tar_x, tar_y, tar_z, loc_x, loc_y, loc_z)
         do k=st_idz,en_idz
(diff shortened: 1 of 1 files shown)


mcbarton (approved):

warrickball:
I know I'm late to this and unfamiliar with OpenACC, but do we not need 
```f90
!$acc end parallel loop
```
corresponding to each loop? Or is it idiomatic to exclude them? I can believe that the compiler can still unambiguously match the end of the parallel block to the end of the loop but I just expected them to be there anyway.

tomneep:
It isn't required, the examples of `parallel loop` in [the spec](https://www.openacc.org/sites/default/files/inline-images/Specification/OpenACC-3.4.pdf) don't include it, and it is optional on L2769 (note this is from a newer spec than most compilers probably support but it is also shown as optional in older specs). Perhaps we can add it for completeness since we have it in plenty of other places.

warrickball:
No problem, thanks for linking the spec. I don't mind omitting it. I couldn't find any sign of consensus online.

Issue #165, opened 2025-09-19
`blk` argument functions for MPI


tomneep:
In the MPI R&D branch, the following subroutines were changed to take a `Block` type as an argument. Not that some of the subroutines are marked with `(warning still requires to know whether block is coarse or not)`, these are just ones where I passed both the block and the block number to the subroutine, but there is certainly a better way of dealing with this that is less hacky.

Note that some care should be taken when moving the loops into `main.F90`, as some of the subroutines only ever act on the either the coarse block or fine blocks.

## AllocateArrays.f90
- [x] allocateArrays #154 

## BoundaryConditions.f90
- [x] velocityBC https://github.com/BioFSILab/Bio-CFD/pull/177
- [x] solidCellBC https://github.com/BioFSILab/Bio-CFD/pull/179
- [x] solidCellBC_move https://github.com/BioFSILab/Bio-CFD/pull/181

## CoefficientMatrix.f90
- [x] coefficientMatrix (warning still requires to know whether block is coarse or not) #207 

## Forcing.f90
- [x] pressureForcing1 https://github.com/BioFSILab/Bio-CFD/pull/187
- [x] pressureforcingfield https://github.com/BioFSILab/Bio-CFD/pull/188
- [x] pressureforcingghost https://github.com/BioFSILab/Bio-CFD/pull/189
- [x] velocityforcing1 https://github.com/BioFSILab/Bio-CFD/pull/190
- [x] velocityforcingfield https://github.com/BioFSILab/Bio-CFD/pull/191
- [x] velocityforcingghost https://github.com/BioFSILab/Bio-CFD/pull/192

## InititalConditions.f90
- [x] initialConditions #203

## LastConditions.f90
- [x] lastConditions #205

## NavierStokes.f90
- [x] non_uni_coeff #201 
- [x] nsmomentum2order #201 

## PcorVcor.F90
- [x] updateVelocity_newv #200

## ReadInput.f90
- [x] readSurfaceMeshGmsh #163 

## Search.F90
- [x] findDistNode #160 
- [x] shiftSurfaceNodesInitial https://github.com/BioFSILab/Bio-CFD/pull/167
- [x] computeSurfaceVariables (warning still requires to know whether block is coarse or not) https://github.com/BioFSILab/Bio-CFD/pull/168
- [x] computeSurfaceNorm https://github.com/BioFSILab/Bio-CFD/pull/170
[...]

PR #167, opened 2025-09-22
Add explicit block argument to shiftSurfaceNodesInitial

mcbarton:
This PR adds explicit block argument to shiftSurfaceNodesInitial

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -39,8 +39,8 @@
         end do
         do g=blk_start, size(block)
            CALL findDistnode(block(g))
+           CALL shiftSurfaceNodesInitial(block(g))
         end do
-        CALL shiftSurfaceNodesInitial
         CALL computeSurfaceNorm
         CALL interfaceDetail
         totalTime=0.
--- a/app/main.F90
+++ b/app/main.F90
@@ -2,7 +2,8 @@
       PROGRAM main
         use, intrinsic :: iso_fortran_env, only: int64, dp => real64
         USE global, only: block, blk_start, coarse_flcnt_check, deltat, istart, &
-             ita, ita1, ita2, itamax, nblocks, totaltime, totime
+             ita, ita1, ita2, itamax, nblocks, totaltime, totime,a0y,alpha_m, &
+             alpha_t,ang_theta,aoa,aoa1,aoa2,phase_angle,pi,theta_m,theta_t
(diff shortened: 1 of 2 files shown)


tomneep:
Hmmm... I see a difference in outputs with this branch. Will double check. Hopefully I've just done something wrong.

app/main.F90
```diff
@@ -40,7 +41,18 @@ PROGRAM main
...
+        phase_angle = phase_angle*pi/180_dp
+        aoa1 = aoa*pi/180_dp
+        aoa2 = -aoa1
+        alpha_m = alpha_m*pi/180_dp
+        theta_m = theta_m*pi/180_dp
+        a0y = 0.  !a0y
+        ang_theta = 0.  !2._dp*pi*freq
+        alpha_t=(alpha_m*0.5_dp)*(1+cos(ang_theta*(totime+deltat)+phase_angle))
+        theta_t       =  theta_m*cos(ang_theta*(totime+deltat))
+        do g=blk_start, size(block)
```
warrickball, line 53: Only potentially meaningful difference I've noticed in a quick glance is that here the loop is up to `size(block)` whereas in the previous `shiftSurfaceNodesInitial` it was up to `nblocks`.

warrickball: Ah, sorry, the differences might have already been fixed while I had this open.

tomneep: Yeah the differences are elsewhere, but just for future reference, `nblocks == size(block)`

app/main.F90
```diff
@@ -40,7 +41,18 @@ PROGRAM main
...
-        CALL shiftSurfaceNodesInitial
+        phase_angle = phase_angle*pi/180_dp
+        aoa1 = aoa*pi/180_dp
+        aoa2 = -aoa1
+        alpha_m = alpha_m*pi/180_dp
+        theta_m = theta_m*pi/180_dp
+        a0y = 0.  !a0y
+        ang_theta = 0.  !2._dp*pi*freq
+        alpha_t=(alpha_m*0.5_dp)*(1+cos(ang_theta*(totime+deltat)+phase_angle))
+        theta_t       =  theta_m*cos(ang_theta*(totime+deltat))
```
tomneep, line 52: I'm fine with this for now but we should probably try and tidy this up in the future, as it isn't so nice to end up with all this clogging up the main program.

tomneep (approved):

PR #172, opened 2025-09-22
Add explicit block argument to tagging_th

mcbarton:
This PR dds explicit block argument to tagging_th

### DIFF
--- a/src/Search.F90
+++ b/src/Search.F90
@@ -264,20 +264,20 @@ module biocfd_search
 
      END SUBROUTINE computeSurfaceNorm
 
-     SUBROUTINE tagging_th
-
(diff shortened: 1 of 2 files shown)


src/Search.F90
```diff
@@ -295,28 +295,28 @@ SUBROUTINE tagging_th
 #ifndef _OPENACC
         !$omp parallel do default(none) private(minDis, minDis1) &
         !$omp& private(dis_cen, dis_pnt, nel2Cen, nel2Pnt, n2dotn) &
-        !$omp& shared(g, block)
+        !$omp& shared(blk_no, blk)
```
tomneep, line 298: I'm going to approve & merge, but I don't think `blk_no` is needed here as it isn't used inside the `omp` region.

tomneep (approved):

PR #174, opened 2025-09-23
Add explicit block argument to findTScells

mcbarton:
This PR adds an explicit block argument to findTScells

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -61,7 +61,6 @@
         IF (iStart==1) CALL lastConditions
         CALL computeNormDistance
         do g=blk_start, size(block)
-           print*,'inside findTScells'
            CALL findTScells(block(g))
         end do
         CALL coefficientMatrix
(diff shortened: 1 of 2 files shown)


app/main.F90
```diff
@@ -60,7 +60,10 @@ PROGRAM main
         IF (iStart==0) CALL initialConditions
         IF (iStart==1) CALL lastConditions
         CALL computeNormDistance
-        CALL findTScells
+        do g=blk_start, size(block)
+           print*,'inside findTScells'
```
tomneep: I'm guessing this is an attempt to keep the output the same? It isn't because in the original subroutine it wasn't inside the loop. I think we can just delete this print statement as it doesn't offer any informative information (unlike others which give us things like cell counts)

mcbarton: Yes, that was an attempt to keep the output the same. Will remove.

mcbarton: Done

tomneep (approved):

tomneep:
Thanks, can be merged once CI passes

PR #182, opened 2025-09-24
Refactor tagging_th_move and tagging_th to call a new tagging_th_core

mcbarton:


tomneep:
I'm fairly sure we do want to do this, I'd just like to see if we can think of someway to test it before merging. The difference with other refactors is that the `_move` versions aren't actually called very often, so you need to run a large number of iterations for it to be called... I am tempted to just put it in though

tomneep (commented):
I don't know how we test this. I'd also be tempted to split deleting the file writes to a different PR. What about a new `tagging_th_core` subroutine that is called by both `tagging_th` and `tagging_th_mv`? And then if we get to the point where `tagging_th == tagging_th_core` then we can just delete/rename one?

tomneep:
Just to be clear I'm definitely in favour of removing the files, and I think we had permission to do this in one of our meetings

src/Search.F90
```diff
@@ -391,137 +374,53 @@ SUBROUTINE tagging_th(blk,blk_no)
...
+           call tagging_th_core(block(g))
         endif
         ENDDO
         DO g=1,intflines
         a_blk_no=intfr(g)%a_blk
         b_blk_no=intfr(g)%b_blk
 
         if ( block(b_blk_no)%move_check == 1)then
-       call fineUpdate_mv(g)
+      call fineUpdate_mv(g)
```
tomneep: I suppose this is an accidental whitespace change?

src/Search.F90
```diff
@@ -391,137 +374,53 @@ SUBROUTINE tagging_th(blk,blk_no)
          END DO
          END DO
 
+         print*, 'search done'
+         Print*, 'imms. cells=', blk%ibCellCount
+        Print*, 'fluid cells=',blk%fluidCellCount
+        Print*, 'solid cells=', blk%solidCellCount
+     END SUBROUTINE tagging_th_core
+
+     SUBROUTINE tagging_th(blk,blk_no)
```
tomneep, line 383: Shouldn't `tagging_th_core` be called in this subroutine? What am I missing...

tomneep (changes requested):

src/Search.F90
mcbarton: Fixed.

src/Search.F90
mcbarton: Yes, it should have. I prepared the subroutine for it, but then forgot to add the call. Great catch. Fixed now.

src/Search.F90
```diff
@@ -391,129 +374,46 @@ SUBROUTINE tagging_th(blk,blk_no)
...
+       INTEGER(int64), intent(in)  :: blk_no
+       CHARACTER(LEN=120) :: filename1
+
+        call tagging_th_core(blk)
+               WRITE(filename1,1) blk_no
+  1      FORMAT('butter_f.',i3.3,".dat")
+          OPEN(11,FILE=filename1,status='unknown')
+        DO k = 1, blk%nz+2
+        DO j = 1, blk%ny+2
+        DO i = 1, blk%nx+2
```
tomneep, line 395: These variables aren't defined

tomneep (approved):

PR #183, opened 2025-09-24
Move u_dum,v_dum,w_dum and p_dum outside of Block and make them local variables in change_block_coords

mcbarton:


src/Search.F90
```diff
@@ -1326,10 +1316,27 @@ SUBROUTINE change_block_coords
         INTEGER(int64) :: i,j,k,g, a_blk_no, b_blk_no,countx_st,countz_st,county_st
         REAL(dp) :: change_y_f,change_x_f
         REAL(dp) :: change_z_f
+        REAL (dp), ALLOCATABLE, DIMENSION (:, :, :) :: u_dum,v_dum,w_dum,p_dum
         DO g=blk_start,nblocks
 
         if ( block(g)% move_check == 1) then
 
+        ALLOCATE(&
```
tomneep, line 1216: These will be allocated every time this subroutine is called so I expect it will fail on a second call?

warrickball: AFAIK [allocatable arrays are automatically de-allocated](https://fortran-lang.org/learn/best_practices/allocatable_arrays/) when they go out of scope.

I think whether or not you want to explicitly de-allocate anyway before exiting the procedure is a matter of style.

tomneep: That is very useful to know! In that case this probably works ok. I'll have another look.

mcbarton: @tomneep Did you have a chance to take another look at this PR?

tomneep (commented):
I'm wondering if you had a chance to test that there is no impact on performance here? I suppose there shouldn't be much, but we go from a single allocation to allocating and deallocating every time this subroutine is called don't we?

mcbarton:
> I'm wondering if you had a chance to test that there is no impact on performance here? I suppose there shouldn't be much, but we go from a single allocation to allocating and deallocating every time this subroutine is called don't we?

I have not done a performance check. I suspect if there is performance penalty, it will be offset by the fact we are no longer constantly moving these rather large arrays back and forth from the gpu for no reason. Happy for a performance benchmark to be performed on the PR to check though.

src/Search.F90
tomneep: I'm still not sure this will work as it is because the allocations are happening in a loop, so on the 2nd iteration won't it try to reallocate the arrays even though they are allocated?

warrickball: Ah, good point. I missed that this is in a loop and I'm with Tom here: this looks like it'll crash. Here's a small example which I think is representative:
```
$ cat -n loop-alloc.f90
     1	program mwe
     2	
     3	  implicit none
     4	
     5	  integer, parameter :: dp = KIND(0.0d0)
     6	
     7	  call alloc_in_loop
     8	
     9	contains
    10	
    11	  subroutine alloc_in_loop
    12	    integer :: i
    13	    real(dp), allocatable :: x(:,:,:)
    14	
    15	    do i = 1,4
    16	      allocate(x(3,3,3))
    17	    end do
    18	  end subroutine alloc_in_loop
    19	
    20	end program mwe
$ gfortran loop-alloc.f90 -o loop-alloc.x
$ ./loop-alloc.x
At line 16 of file loop-alloc.f90
Fortran runtime error: Attempting to allocate already allocated variable 'x'

Error termination. Backtrace:
#0  0x7f9268209055 in ???
#1  0x7f9268209c19 in ???
#2  0x7f926820a1a9 in ???
#3  0x400579 in ???
#4  0x400646 in ???
#5  0x40067c in ???
#6  0x7f9268011574 in ???
#7  0x7f9268011627 in ???
#8  0x400404 in ???
#9  0xffffffffffffffff in ???
```
Adding `deallocate(x)` between lines 16 and 17 prevents the error.

mcbarton: I have now added the necessary deallocate statement

tomneep:
I've run a long test on Baskerville to try and test this. I get an error after 435 iterations (exit code 135). On the dev branch it carries on without problems and shows that iteration 436 is where `change_block_coords` is actually called. It could be a coincidence but it seems like this PR is likely responsible for the error.

PR #184, opened 2025-09-24
Remove unused variables

mcbarton:
These variables appear to be unused

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -29,7 +29,6 @@
         IMPLICIT NONE
 
         INTEGER (int64) :: g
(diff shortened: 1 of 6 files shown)


tomneep:
I get a runtime error. 

```
At line 34 of file ././src/ReadInput.f90
Fortran runtime error: Cannot match namelist object name v0
```

tomneep:
Removing these from the namelist file fixes this. Will need to update https://github.com/BioFSILab/biocfd-test-input

tomneep (approved):

PR #196, opened 2025-10-01
Remove block, blk_start, nblocks from use global BoundaryConditions.f90

mcbarton:
Now blk is an argument of all the subroutines in BoundaryConditions.f90 we can remove block, blk_start, nblocks from use global in BoundaryConditions.f90


tomneep:
It is a bit disappointing this isn't picked up by the CI

tomneep (approved):

PR #199, opened 2025-10-01
Remove use Global from BoundaryConditions.f90

mcbarton:
Now that blk has been made an argument of the subroutines in BoundaryConditions.f90, it is not too much effort to remove the few variables still being used in the Global module from BoundaryConditions.f90, so this PR removes them.


mcbarton:
@warrickball @tomneep pining for review.

tomneep (approved):

PR #200, opened 2025-10-02
Add blk argument to updateVelocity_newv

tomneep:
Also made the `openacc` directive collapse all three loops (because I'm not sure there is any downside to asking that) and added a unit test.

I'm not really sure why this subroutine is in `PcorVcor`. It seems like it could be somewhere else...

Part of #165

### DIFF
--- a/src/PcorVcor.F90
+++ b/src/PcorVcor.F90
@@ -3,6 +3,7 @@ module biocfd_pcor_vcor
   use global, only : block, deltat, epsi, omega, omega1, omega2, omega3, omega4, &
        ita, nblocks, totaltime, &
        totime
+  use biocfd_blocks, only: Blocks
 #ifdef _OPENMP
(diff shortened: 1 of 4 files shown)


src/PcorVcor.F90
```diff
@@ -3,6 +3,7 @@ module biocfd_pcor_vcor
   use global, only : block, deltat, epsi, omega, omega1, omega2, omega3, omega4, pcitamax, &
```
mcbarton:
```suggestion
  use global, only : deltat, epsi, omega, omega1, omega2, omega3, omega4, pcitamax, &
```

tomneep: We can't remove `block` here yet as it is used in other subroutines, I just didn't need to change these for the MPI branch.

mcbarton (approved):

PR #201, opened 2025-10-02
Add blk argument to both Navier Stokes subroutines

tomneep:
Add explicit block argument to both Navier-Stokes subroutines. Probably should be two PRs so let me know if you want me to split.

Progress towards #165

### DIFF
--- a/src/NavierStokes.f90
+++ b/src/NavierStokes.f90
@@ -12,7 +12,7 @@ contains
   SUBROUTINE non_uni_coeff(blk)
 
        type(Blocks), intent(inout) :: blk
-       INTEGER  (dp) :: i, j, k, g, nx_var, ny_var, nz_var
+       INTEGER  (dp) :: i, j, k, nx_var, ny_var, nz_var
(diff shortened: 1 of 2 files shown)


mcbarton (approved):

mcbarton:
@tomneep Once you fix the merge conflict, feel free to merge.

src/NavierStokes.f90
```diff
@@ -1,15 +1,18 @@
 module biocfd_navier_stokes
   use, intrinsic :: iso_fortran_env, only: dp => real64, int64
-  use global, only : block, alpha, deltat, nblocks, re
+  use global, only : alpha, deltat, nblocks, re
```
mcbarton:
```suggestion
  use global, only : alpha, deltat, re
```

We are now no longer using nblocks in NavierStokes.f90

tomneep: Fixed. Might conflict with #217

mcbarton (approved):

PR #203, opened 2025-10-02
Add blk argument to InitialConditions

tomneep:
For #165.

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -3,7 +3,7 @@
         use, intrinsic :: iso_fortran_env, only: int64, dp => real64
         USE global, only: block, blk_start, coarse_flcnt_check, deltat, istart, &
              ita, ita1, ita2, itamax, nblocks, totaltime, totime,alpha_m, &
-             ang_theta,aoa,aoa1,aoa2,phase_angle,pi,theta_m
+             ang_theta,aoa,aoa1,aoa2,phase_angle,pi,theta_m, uc
         use biocfd_search, only: findDistnode, shiftSurfaceNodesInitial, computeSurfaceNorm, &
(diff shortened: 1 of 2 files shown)


app/main.F90
```diff
@@ -67,7 +67,16 @@ PROGRAM main
         print*,'13'
         CALL cellCount_solid_coarse(block(1))
         print*,'14'
-        IF (iStart==0) CALL initialConditions
+        IF (iStart==0) then
+           WRITE(*,*) 'Enter initialcondtitions'
+           ita = 0
+           ita1 = 0
+           totime = 0.
```
tomneep, line 74: For now I'm moving this here, although it is a bit messy. Certainly we can probably ditch the write/print statements at some point in the future

src/InitialConditions.f90
```diff
@@ -1,48 +1,38 @@
...
+    blk%vt = 0._dp
+    blk%w = 0._dp
+    blk%wt = 0._dp
+    blk%p = 0._dp
+    blk%cell_pr=0
+    DO n = 1, blk%fluidCellCount
+       i = blk%fluidIndexPtr(n, 1)
+       j = blk%fluidIndexPtr(n, 2)
+       k = blk%fluidIndexPtr(n, 3)
+       blk%u(i,j,k) = uc  !396.33054782262406 !116.236233
```
tomneep: Might be the time to delete these?
```suggestion
       blk%u(i,j,k) = uc
```

app/main.F90
mcbarton: I am fine with this. We can remove all these random print statements after we start combining loops in main. I suggest we just do one PR where we remove them all.

mcbarton (approved):

PR #205, opened 2025-10-02
Use explicit blk argument in LastConditions

tomneep:
For #165

### DIFF
--- a/src/LastConditions.f90
+++ b/src/LastConditions.f90
@@ -1,52 +1,50 @@
 module biocfd_last_conditions
-  use, intrinsic :: iso_fortran_env, only : dp => real64
-  use global, only : block, ita, ita1, nblocks, re, totime
+  use, intrinsic :: iso_fortran_env, only : dp => real64, int64
+  use global, only: ita, ita1, totime
+  use biocfd_blocks, only: Blocks
   implicit none
   private
 
   public :: lastConditions
 
-  contains
-      SUBROUTINE lastConditions
+contains
+  SUBROUTINE lastConditions(blk, id, re)
 
-       INTEGER::  i, j, k, g
-        CHARACTER(len=150) :: filename3
-        WRITE(*,*) 'Enter lastcondtitions'
+    !> The Block that we want to setup
+    type(Blocks), intent(inout) :: blk
+    !> The ID number of the block, usually 1 for the coarse block and
+    !> >=2 for the fine blocks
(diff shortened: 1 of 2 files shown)


src/LastConditions.f90
```diff
@@ -1,53 +1,50 @@
...
+    type(Blocks), intent(inout) :: blk
+    !> The ID number of the block, usually 1 for the coarse block and
+    !> >=2 for the fine blocks
+    integer(int64), intent(in) :: id
+    real(dp), intent(in) :: re
+    INTEGER ::  i, j, k
+    CHARACTER(len=150) :: filename3
+
+    WRITE(filename3,3) id, re
+3   FORMAT('out/aorta_chkpt.',i3.3,'.',f6.1,".dat")
```
tomneep, line 23: This is the wrong filename, but there are probably bigger issues to address with checkpointing. I'll open an issue.

tomneep: #206

src/LastConditions.f90
```diff
@@ -1,53 +1,50 @@
...
+    OPEN (1, FILE=filename3, FORM='formatted')
+    DO k = 1, blk%nz+2
+       DO j = 1, blk%ny+2
+          DO i = 1, blk%nx+2
+             ! TODO: totime, ita, and ita1 are all global variables,
+             ! but they will always just be overwritten by the last
+             ! block which is read. Do we want to have some validation
+             ! here that things are working as one would expect?
+             READ(1,*) blk%u(i,j,k), blk%v(i,j,k), blk%w(i,j,k), blk%p(i,j,k), &
+                  totime, ita, ita1
```
warrickball, line 33: I think that depends on if they're used again after this routine, or if they're just used here to match the format of the line because it's user-facing data the code doesn't need to restart.

A crude way of checking this would be to invent some local variables of the same type, use those instead, and see if the output changes. I would *hope* not.

warrickball: If they are just discarded, I'd at least leave a comment like "the last three variables on the line are blah but we don't need them".

tomneep: Pretty sure that `ita` and `totime` are crucial. `ita1` can probably be removed from the code, I don't think it is doing anything useful.

mcbarton:
@tomneep can you fix the merge conflicts for this PR?

mcbarton (approved):

PR #207, opened 2025-10-02
Explicit block argument in coefficient matrix

tomneep:
For #165

### DIFF
--- a/src/CoefficientMatrix.f90
+++ b/src/CoefficientMatrix.f90
@@ -1,73 +1,72 @@
 module biocfd_coefficient_matrix
   use, intrinsic :: iso_fortran_env, only: dp => real64, int64
-  use global, only : block, nblocks
+  use biocfd_blocks, only: Blocks
   implicit NONE
   private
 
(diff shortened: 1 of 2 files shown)


src/CoefficientMatrix.f90
```diff
@@ -1,73 +1,72 @@
...
-  contains
-SUBROUTINE coefficientMatrix
+contains
+  SUBROUTINE coefficientMatrix(blk, is_coarse)
 
-        INTEGER (int64) :: i, j, f,nx_var,ny_var,nz_var
-        REAL (dp)    ::  rx1, rx2, rxsum, ry1, ry2, rysum, rz1, rz2, rzsum
+    !> The block to construct the coefficient matrix for
+    type(Blocks), intent(inout) :: blk
+    logical, intent(in) :: is_coarse
```
tomneep, line 14: Rather than passing in the ID of the block, I decided here it is a bit nicer to pass a `logical` of whether the block is coarse or not.

mcbarton (approved):

PR #208, opened 2025-10-02
Blk argument cell count solid

tomneep:
For #165. 

There are two commits here. The first is the one to add the `blk` argument. The second does some minor formatting clean up (mostly just selecting the subroutine in emacs and pressing tab).

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -61,7 +61,11 @@
            CALL tagging_th(block(g),g)
         end do
         print*,'11'
-        CALL cellCount_solid
+        print*, "cellCount started"
+        do g=blk_start, size(block)
+           CALL cellCount_solid(block(g))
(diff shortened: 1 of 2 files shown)


app/main.F90
```diff
@@ -142,8 +146,11 @@ PROGRAM main
              block(g)% u1_ghost,block(g)% u1t_ghost,block(g)% v1_ghost,block(g)% v1t_ghost, &
              block(g)%w1_ghost,block(g)% w1t_ghost)
        END DO
-
-            CALL cellCount_solid
+       print*, "cellCount started"
+       do g=blk_start, size(block)
+          CALL cellCount_solid(block(g))
+          print*,g, block(g)%fluidCellCount, block(g)%redCellCount, block(g)%blackCellCount
```
tomneep, line 152: I moved this `print` here for consistent outputs, but probably can remove this in the future

mcbarton (approved):

PR #211, opened 2025-10-03
Make xfact a module parameter rather than global variable

tomneep:
While looking at other things I noticed that `xfact` is:
- Only used in `src/Search.F90`
- Is never changed after being initial set
- Is set by hand i.e. not read from the input files

So I think it makes sense to make this into a `parameter` of the search module. What do you think?


mcbarton (approved):
I am pretty sure that alpha in modGlobal could have a similar change too. Its set to 1.0 in readinput, and then used only in NavierStokes.f90

PR #213, opened 2025-10-07
Remove redundant 0._dp + from Forcing.f90

mcbarton:


src/Forcing.f90
```diff
@@ -118,11 +118,11 @@ SUBROUTINE velocityForcing1(blk)
       DO n = 1, blk%ibCellCount
 
          IF (blk%ibSurfID(blk%nelu2(n))==50) THEN
-             usurf = 0._dp + blk%xdot
+             usurf = blk%xdot
```
mcbarton, line 121: usurf is always set to blk%xdot for  blk%ibSurfID(blk%nelu2(n))=50,51,52 . Its undefined if not one of these values. May be worth removing if statements, and just have a line which sets usurf to blk%xdot, or replace every occurrence of usurf with blk%xdot since xdot is not redefined in Forcing.f90

tomneep (approved):

PR #214, opened 2025-10-07
Remove variables from global

mcbarton:


src/ReadInput.f90
```diff
@@ -28,7 +29,7 @@ SUBROUTINE readInput
                    re,rho_f, mu_f, l_c, &
                    u0,  &
                    surGeoPoints, phase_angle, freq, aoa, piv_pt,alpha_m, theta_m, &
-                   istart, dt_order, inor, dxmin
+                   istart, inor, dxmin
```
mcbarton: dt_order needs removing from the input file. Its not being used.

src/ReadInput.f90
```diff
@@ -91,14 +94,12 @@ SUBROUTINE readInput
         mu_f = mu_f*1e-5_dp
         rho_f = rho_f
         l_c = 0.001_dp*l_c
-        rev = rho_f/(mu_f)
-        u0 = re/(rev*l_c)
+        u0 = re*mu_f/(rho_f*l_c)
         u0 = u0*1
         uc=u0*1
```
warrickball: Is this not the same as
```suggestion
        uc = u0
```
? Presumably `*1` does nothing and neither does `u0 = u0`?

mcbarton:
@tomneep @warrickball This PR is ready for review. There are many variables left in modGlobal, but this takes care of a good chunk of them. I'll remove the rest in similar subsequent PRs over the next few months.

src/ReadInput.f90
tomneep: Link to PR in https://github.com/BioFSILab/biocfd-test-input?

tomneep: Actually I think we need to keep `dt_order`

tomneep (approved):

Issue #194, opened 2025-10-01
Better type names


tomneep:
I think we currently have two user defined types: `Blocks` and `Interfaces`. I think these should be renamed to at least be singular and not plural, since each time we make an instance of one of these we are only defining one of them e.g.
```f90
type(Blocks) :: single_block
type(Interfaces) :: single_interface
```
I think there is a convention to use `_t` when defining types, so I think we should have
```f90
type(Block_t) :: single_block
type(Interface_t) :: single_interface
```
but happy to hear other proposals

warrickball:
Not sure if there's consensus, even in (subjectively) high-profile and high-quality projects. I've definitely seen the `_t` convention, e.g. [in GYRE](https://github.com/rhdtownsend/gyre/blob/master/src/nad/nad_bvp_m.fypp). But equally I noticed that the Fortran stdlib project [uses `_type`](https://github.com/fortran-lang/stdlib/blob/master/src/stdlib_linalg_state.fypp) but `fpm` itself [doesn't use anything](https://github.com/fortran-lang/fpm/blob/main/src/fpm_command_line.f90).

FWIW, personally I'm in favour of the `_t` affix.

PR #215, opened 2025-10-10
Rename Interfaces to Interface_t

tomneep:
Half of #194 (presuming we are in agreement about it)


mcbarton (approved):
This new name for the interface type makes sense to me.

PR #218, opened 2025-10-10
Make blk argument of write_output_ascii

mcbarton:
This PR makes blk argument of write_output_ascii

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -110,7 +110,7 @@
         CALL write_output_hdf5
 #else
         do g=1, size(block)
-           CALL write_output_ascii(block(g),g,block(2)%dx,char_f)
+           CALL write_output_ascii(block(g),g,char_f)
         end do
 #endif
         coarse_flcnt_check=0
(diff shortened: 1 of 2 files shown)


app/main.F90
```diff
@@ -102,7 +102,9 @@ PROGRAM main
 #if USE_HDF5 == 1
         CALL write_output_hdf5
 #else
-        CALL write_output_ascii(char_f)
+        do g=1, size(block)
+           CALL write_output_ascii(block(g),g,block(2)%dx,char_f)
```
mcbarton: I didn't like having to put block(2)%dx as an argument of the subroutine, but for some reason the filename is based on it

tomneep: That must be a bug... what happens in the single butterfly case?

mcbarton: I can either remove dx from the filename, or change the filename to be based on blk%dx. Which would you think would be best to go with it?

mcbarton: Fixed by changing filename to be based on blk%dx

tomneep (approved):

PR #219, opened 2025-10-10
Move alpha_m, theta_m, alpha_m1, theta_m1, mu_f, rho_f, l_c, u_tip, disp out of modGlobal.f90 into readInput subroutine

mcbarton:


mcbarton:
Alot of these variables do nothing useful, but kept them in for consistency of output.

src/ReadInput.f90
```diff
@@ -26,7 +26,7 @@ SUBROUTINE readInput(surGeoPoints,char_f,istart,itamax,pcItaMax)
        !     Variables removed from Blocks 'xstart, xend, ystart, yend, zstart, zend'
        REAL(dp),ALLOCATABLE,DIMENSION(:) :: xstart_temp,xend_temp,&
        ystart_temp,yend_temp,zstart_temp,zend_temp
-
+       REAL(dp) :: alpha_m,theta_m,alpha_m1,theta_m1,mu_f,rho_f,l_c,u_tip,disp
```
tomneep, line 29: Since `alpha_m1` and `theta_m1` are just set to `abs(alpha_m)` (and `abs(theta_m)`) and then written out, how about we just directly write out `abs(..._m)` and then we can delete the `_m1` versions?

tomneep (approved):
Just one comment that you may or may not want to add to this PR. Otherwise looks good to me.

PR #220, opened 2025-10-10
Prepare change_block_coords for MPI

mcbarton:

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -28,7 +28,7 @@
              velocityforcing1, velocityforcingfield, velocityforcingghost
         IMPLICIT NONE
 
-        INTEGER (int64) :: g
+        INTEGER (int64) :: g,m
         INTEGER (int64)   :: surGeoPoints
         CHARACTER (LEN = 3)   :: char_f
         INTEGER               :: istart
@@ -154,9 +154,9 @@
            CALL block_move_check
            DO g=blk_start, size(block)
               CALL change_block_coords(block(g))
-           END DO
-           DO g=1,intflines
-              CALL change_block_coords_interfaces(intfr(g),block(intfr(g)%b_blk))
+              DO m=1,intflines
+                CALL change_block_coords_interfaces(intfr(g),block(intfr(g)%b_blk))
+              END DO
            END DO
            CALL change_block_interface
           CALL fine_block_cell
--- a/app/main.F90
+++ b/app/main.F90
@@ -3,7 +3,7 @@
         use, intrinsic :: iso_fortran_env, only: int64, dp => real64
         USE global, only: block, blk_start, coarse_flcnt_check, deltat, &
              ita, ita1, nblocks, totaltime, totime,alpha_m, &
-             aoa,aoa1,aoa2,phase_angle,pi,theta_m, uc, re, intfr, intflines
+             aoa,aoa1,aoa2,phase_angle,pi,theta_m, uc, re, intfr
         use biocfd_search, only: findDistnode, shiftSurfaceNodesInitial, computeSurfaceNorm, &
              tagging_th, tagging_th_move, block_move_check, cellcount_solid, &
              cellcount_solid_coarse, cellcount_solid_coarse_mv, change_block_coords, &
@@ -28,7 +28,7 @@
              velocityforcing1, velocityforcingfield, velocityforcingghost
         IMPLICIT NONE
 
-        INTEGER (int64) :: g,m
+        INTEGER (int64) :: g
         INTEGER (int64)   :: surGeoPoints
         CHARACTER (LEN = 3)   :: char_f
         INTEGER               :: istart
@@ -154,9 +154,9 @@
            CALL block_move_check
            DO g=blk_start, size(block)
               CALL change_block_coords(block(g))
-              DO m=1,intflines
-                CALL change_block_coords_interfaces(intfr(g),block(intfr(g)%b_blk))
-              END DO
+           END DO
+           DO g=1,size(intfr)
+              CALL change_block_coords_interfaces(intfr(g),block(intfr(g)%b_blk))
            END DO
            CALL change_block_interface
           CALL fine_block_cell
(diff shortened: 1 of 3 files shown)


app/main.F90
```diff
@@ -3,12 +3,12 @@ PROGRAM main
         use, intrinsic :: iso_fortran_env, only: int64, dp => real64
         USE global, only: block, blk_start, coarse_flcnt_check, deltat, &
              ita, ita1, nblocks, totaltime, totime,alpha_m, &
-             aoa,aoa1,aoa2,phase_angle,pi,theta_m, uc, re
+             aoa,aoa1,aoa2,phase_angle,pi,theta_m, uc, re, intfr, intflines
```
tomneep: Use `size(intfr)` instead?

app/main.F90
```diff
@@ -28,7 +28,7 @@ PROGRAM main
              velocityforcing1, velocityforcingfield, velocityforcingghost
         IMPLICIT NONE
 
-        INTEGER (int64) :: g
+        INTEGER (int64) :: g,m
```
tomneep: Nit pick for the future, generally I think it nice to have spaces after commas

app/main.F90
```diff
@@ -152,7 +152,12 @@ PROGRAM main
            CALL computeSurfaceVariables(block(g),g)
         END DO
            CALL block_move_check
-           CALL change_block_coords
+           DO g=blk_start, size(block)
+              CALL change_block_coords(block(g))
+              DO m=1,intflines
```
tomneep: Is having the nested loops right? Looking at the existing subroutine it looks like the block loop happens and the interface loop? I might be mistaken...

src/Interfaces.f90
```diff
@@ -13,10 +13,11 @@ MODULE biocfd_interface_type
                                                         vz_interface_det, wz_interface_det, &
                                                         py_interface_det, uy_interface_det, &
                                                         vy_interface_det, wy_interface_det
-        REAL(dp) :: a_blk, b_blk, xintf_start, xintf_end, yintf_start, yintf_end, &
+        REAL(dp) :: a_blk, xintf_start, xintf_end, yintf_start, yintf_end, &
                     zintf_start, zintf_end
         REAL(dp) :: xintf_st_new, xintf_en_new, yintf_st_new, yintf_en_new
         REAL(dp) :: zintf_st_new, zintf_en_new
+        INTEGER(int64) :: b_blk
```
tomneep: Can this just be a plain `integer`? Can `a_blk` be changed now too?

src/Search.F90
```diff
@@ -1308,170 +1309,168 @@ SUBROUTINE block_move_check
...
-        if( block(b_blk_no)%xp(i) > (intfr(g)%xintf_st_new  + block(b_blk_no)%dx) .and. &
-               block(b_blk_no)%xp(i) < (intfr(g)%xintf_en_new  - block(b_blk_no)%dx).and. &
-               block(b_blk_no)%zp(k) > (intfr(g)%zintf_st_new  + block(b_blk_no)%dz) .and. &
-               block(b_blk_no)%zp(k) < (intfr(g)%zintf_en_new  - block(b_blk_no)%dz).and. &
-               block(b_blk_no)%yp(j) > (intfr(g)%yintf_st_new  + block(b_blk_no)%dy).and. &
-               block(b_blk_no)%yp(j) < (intfr(g)%yintf_en_new  - block(b_blk_no)%dy))then
+      end subroutine change_block_coords
 
-                block(b_blk_no)%cell_n(i,j,k)=1
+      subroutine change_block_coords_interfaces(local_intfr,blk)
```
tomneep, line 1282: I know you didn't introduce this notation, but why `intfr`?

src/Search.F90
```diff
@@ -1308,170 +1309,168 @@ SUBROUTINE block_move_check
...
+        type(Interface_t),intent(inout) :: local_intfr
+        type(Blocks), intent(inout) :: blk
+        INTEGER(int64) :: k,j,i,countx_st,countz_st,county_st
+
+        if ( blk% move_check == 1) then
```
tomneep: I generally don't like `if` statements spanning an entire function (I'd do an early return) but I think this is possibly personal preference

src/Search.F90
```diff
@@ -1308,170 +1309,168 @@ SUBROUTINE block_move_check
...
+                  exit
+              endif
+           ENDDO
 
+           print*,'bef'
+           countx_st=blk%cpy_x_start
+           county_st=blk%cpy_y_start
+           countz_st=blk%cpy_z_start
+
+           OPEN(UNIT=12,FILE='log.dat',STATUS='unknown',POSITION='APPEND')
```
tomneep: Ah here is that pointless file. Let's try to remember to delete this!

app/main.F90
tomneep: (and the comment above about using `size(intfr)`)

src/Search.F90
tomneep, line 1282: Let's also write some kind of docstring saying that this has to be called immediately after `change_block_coords` (and that they must be called together)?

tomneep (commented):
This looks to me like a sensible idea. I just have a few comments on small things we can adjust which I hope will make this clearer.

src/Search.F90
```diff
@@ -1214,170 +1215,167 @@ SUBROUTINE block_move_check
...
-               block(b_blk_no)%yp(j) < (intfr(g)%yintf_en_new  - block(b_blk_no)%dy))then
-
-                block(b_blk_no)%cell_n(i,j,k)=1
+      end subroutine change_block_coords
 
+      ! This subroutine used to be combined with change_block_coords
+      ! and as a result this subroutine must be called straight
+      ! after change_block_coords
+      subroutine change_block_coords_interfaces(local_intfr,blk)
+        type(Interface_t),intent(inout) :: local_intfr
```
tomneep: Just to double check, is `local_intfr` actually modified? I might be missing where it is... If not then I think we should set the intent as `in`?

mcbarton: This is something I had missed. Will change once you have finished your checks, so that we don't waste minutes.

tomneep:
I'm seeing output differences, even for just 1 iteration, which is weird. Investigating.

tomneep:
> I'm seeing output differences, even for just 1 iteration, which is weird. Investigating.

Ah I think it is just the changes of `a_blk` and `b_blk` to integers...

tomneep (approved):

PR #221, opened 2025-10-10
Refactor change block interface

mcbarton:
The purpose of this PR is to refactor change block interface, by rewriting the do loops which run over elements as a single line, and utilise print_interface_detail to print the details of the interface. The motivation of this change is to reduce the overall number of lines in the code.

### DIFF
--- a/src/Search.F90
+++ b/src/Search.F90
@@ -6,6 +6,7 @@ module biocfd_search
        coarse_flcnt_check, intfr
   use biocfd_fine_interp, only: fineUpdate_mv
   use biocfd_fine_interp_bound, only : fineUpdate_bd_mv
+  use biocfd_interface_detail, only : print_interface_detail
   use biocfd_block_type, only: Block_t
   implicit NONE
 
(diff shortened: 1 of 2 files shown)


tomneep:
There are differences in the output that need investigating. Weirdly it seems to be the `y` components of just the first interface...

I can't see why that would be any different though...

src/Search.F90
```diff
@@ -1382,251 +1383,104 @@ SUBROUTINE change_block_coords
...
+SUBROUTINE change_block_interface
+
+  INTEGER(int64) :: j,g, a_blk_no, b_blk_no, factor,increment
+
+  do g=1,size(intfr)
+     b_blk_no=intfr(g)%b_blk
+     if ( block(b_blk_no)% move_check /= 1) cycle
+     a_blk_no=intfr(g)%a_blk
+     factor=intfr(g)%b_msh/intfr(g)%a_msh
+     increment = block(b_blk_no)%move_amtx/factor
```
tomneep: Ah...increment is always using `move_amtx`. It should be `move_amty` and `move_amtz` where appropriate

tomneep (changes requested):
See comment about using `move_amtx` everywhere. I think that there should be `move_amty` and `move_amtz`...

src/Search.F90
mcbarton: Done.

src/Search.F90
```diff
@@ -1382,251 +1383,104 @@ SUBROUTINE change_block_coords
...
+
+     intfr(g)%px_interface_det(1,1:intfr(g)%counterxp) =&
```
tomneep:
```suggestion

     increment = block(b_blk_no)%move_amtx/factor
     intfr(g)%px_interface_det(1,1:intfr(g)%counterxp) =&
```

src/Search.F90
```diff
@@ -1382,251 +1383,104 @@ SUBROUTINE change_block_coords
...
+
+     intfr(g)%ux_interface_det(1,1:intfr(g)%counterxu) =&
+             intfr(g)%ux_interface_det(1,1:intfr(g)%counterxu) + increment
+
+     intfr(g)%vx_interface_det(1,1:intfr(g)%counterxv) =&
+             intfr(g)%vx_interface_det(1,1:intfr(g)%counterxv) + increment
+
+     intfr(g)%wx_interface_det(1,1:intfr(g)%counterxw) =&
+             intfr(g)%wx_interface_det(1,1:intfr(g)%counterxw) + increment
+     increment = block(b_blk_no)%move_amty/factor
```
tomneep, line 1409:
```suggestion

     increment = block(b_blk_no)%move_amty/factor
```

src/Search.F90
```diff
@@ -1382,251 +1383,104 @@ SUBROUTINE change_block_coords
...
+
+     intfr(g)%uy_interface_det(1,1:intfr(g)%counteryu) =&
+             intfr(g)%uy_interface_det(1,1:intfr(g)%counteryu) + increment
+
+     intfr(g)%vy_interface_det(1,1:intfr(g)%counteryv) =&
+             intfr(g)%vy_interface_det(1,1:intfr(g)%counteryv) + increment
+
+     intfr(g)%wy_interface_det(1,1:intfr(g)%counteryw) =&
+             intfr(g)%wy_interface_det(1,1:intfr(g)%counteryw) + increment
+     increment = block(b_blk_no)%move_amtz/factor
```
tomneep, line 1422:
```suggestion

     increment = block(b_blk_no)%move_amtz/factor
```

tomneep (commented):
Introduce a blank line between each block so it is easier to see at first glance what is happening (I found the `increment` lines to be a bit hidden when I first looked).

tomneep (approved):
Long tests on Baskerville look good. Approving and merging!

PR #222, opened 2025-10-13
Remove fineUpdate_bd from code

mcbarton:
fineUpdate_bd is just a loop over fineUpdate_bd_mv, so this PR removes it in the code, and replaces it with the loop

### DIFF
--- a/src/PcorVcor.F90
+++ b/src/PcorVcor.F90
@@ -189,7 +189,7 @@ module biocfd_pcor_vcor
          END DO
         !$acc end parallel loop
          END DO
-        DO g=1,intflines
+        DO g=1,size(intfr)
            call fineUpdate_bd_mv(g)
         ENDDO
         CALL coarseUpdate
@@ -374,4 +374,5 @@ module biocfd_pcor_vcor
 
       END SUBROUTINE updateVelocity_newv
 end module biocfd_pcor_vcor
+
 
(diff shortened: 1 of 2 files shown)


tomneep:
For reference #113

src/PcorVcor.F90
```diff
@@ -189,7 +189,9 @@ SUBROUTINE poissonSolver(pcItaMax)
          END DO
         !$acc end parallel loop
          END DO
-        CALL fineUpdate_bd
+        DO g=1,intflines
```
tomneep: I think this has been remove from the code now in #232? Will need to add `intfr` to use global?

mcbarton: Fixed PR.

src/PcorVcor.F90
```diff
@@ -1,15 +1,15 @@
...
+       ita, totaltime,totime,uc, intfr
   use biocfd_block_type, only: Block_t
 #ifdef _OPENMP
   use omp_lib, only: omp_get_max_threads, omp_get_thread_num
 #endif
 #ifdef _OPENACC
   use openacc, only: acc_device_default, acc_get_num_devices, acc_set_device_num
 #endif
-  use biocfd_fine_interp_bound, only : fineUpdate_newv_bd, fineUpdate_bd, fineUpdate_pc_bd
+  use biocfd_fine_interp_bound, only : fineUpdate_newv_bd, fineUpdate_pc_bd,fineupdate_bd_mv
```
tomneep:
```suggestion
  use biocfd_fine_interp_bound, only : fineUpdate_newv_bd, fineUpdate_pc_bd, fineupdate_bd_mv
```

tomneep (approved):

PR #223, opened 2025-10-13
Give fineUpdate_pc_bd input blocks and interface

mcbarton:


tomneep:
I think with MPI now it makes it difficult to do it this way (though not impossible). Perhaps we can close this for now?

mcbarton:
> I think with MPI now it makes it difficult to do it this way (though not impossible). Perhaps we can close this for now?

Happy to close this PR, purely on the basis that with the time we have left on the project I wouldn't be going back to this PR anyway.

PR #225, opened 2025-10-14
Do fine_block_cell on GPU

tomneep:
I don't see any reason that this can't be done. We are just setting variables to 1 if some conditions are met.

### DIFF
--- a/src/Search.F90
+++ b/src/Search.F90
@@ -739,7 +739,7 @@ blk%fluidCellCount = flcnt
             blk%blackCellIndexPtr(blk%blackCellCount,3))
        ipt1 = 0
        iPt = 0
-       !$acc parallel loop
+       !$acc parallel loop private(idx)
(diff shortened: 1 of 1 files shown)


tomneep:
Marking this as ready, tested on Baskerville and no differences in `stdout` other than those reported in #107

src/Search.F90
```diff
@@ -739,7 +739,7 @@ SUBROUTINE cellCount_solid(blk)
             blk%blackCellIndexPtr(blk%blackCellCount,3))
        ipt1 = 0
        iPt = 0
-       !$acc parallel loop
+       !$acc parallel loop private(idx)
```
mcbarton, line 742: This change is not in fine_block_cell. Why is this needed? I see you have a specific commit about this change.

mcbarton (approved):

PR #227, opened 2025-10-17
Make block an argument of write_output_hdf5

mcbarton:

### DIFF
--- a/src/WriteOutput_corner1.F90
+++ b/src/WriteOutput_corner1.F90
@@ -20,7 +20,7 @@ contains
 
 #if USE_HDF5 == 1
       SUBROUTINE write_output_hdf5(blk,blk_no)
-       type(Blocks), intent(in) :: blk
+       type(Block_t), intent(in) :: blk
        integer (int64), intent(in) :: blk_no
        CHARACTER(len=150)  :: filename1
        INTEGER  :: k, i, j
--- a/src/WriteOutput_corner1.F90
+++ b/src/WriteOutput_corner1.F90
@@ -28,7 +28,7 @@ contains
        character (len=11) :: dummy_1
        character (len=5) ::dummy_2
 
-         IF((mod(ita,200_int64) /=0 .or. ita /= 1 .or. ita /=2 )) return 
+         if (mod(ita,200_int64) /=0 .and. ita > 2) return
             write(dummy_1,'(A6,I5.5)') 'block_',blk_no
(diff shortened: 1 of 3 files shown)


src/WriteOutput_corner1.F90
```diff
@@ -19,26 +19,27 @@ module biocfd_write_output_corner1
...
-       INTEGER  :: k, i, j, g
+       INTEGER  :: k, i, j
        REAL (dp), allocatable :: u1(:,:,:), v1(:,:,:), w1(:,:,:)
        character (len=11) :: dummy_1
        character (len=5) ::dummy_2
 
-         IF((mod(ita,200_int64) ==0 .or. ita <= 2 ))then
-         do g=1,nblocks
-            write(dummy_1,'(A6,I5.5)') 'block_',g
+         IF((mod(ita,200_int64) ==0 .or. ita <= 2 )) return
```
tomneep: Did you want to invert the conditions here?

src/WriteOutput_corner1.F90
```diff
@@ -50,30 +51,29 @@ SUBROUTINE write_output_hdf5
...
-                                array_input_3d=block(g)%cell,key='cell',group=dummy_1)
+                                array_input_3d=blk%cell,key='cell',group=dummy_1)
             call hdf5_write_int(filename=filename1,&
-                                array_input_3d=block(g)%cell_n,key='cell_n',group=dummy_1)
+                                array_input_3d=blk%cell_n,key='cell_n',group=dummy_1)
             call hdf5_write_int(filename=filename1,&
-                                array_input_3d=block(g)%cell_pr,key='cell_pr',group=dummy_1)
+                                array_input_3d=blk%cell_pr,key='cell_pr',group=dummy_1)
             deallocate(u1,v1,w1)
          end do
```
tomneep: You've removed `end if` when I think you wanted to remove `end do`?

tomneep (changes requested):
Doesn't compile

src/WriteOutput_corner1.F90
mcbarton: I was supposed to remove both, but forgot to. The end if is removed due to the subroutine doing an early return now.

src/WriteOutput_corner1.F90
mcbarton: Done

tomneep (approved):

Issue #206, opened 2025-10-02
Checkpointing


tomneep:
Checkpointing doesn't work at the moment. This definitely needs to be added before we do any serious running. It probably makes a lot of sense to write checkpoint files as HDF5 files.

To clarify. `LastConditions` (which is the subroutine I presume tries to read the checkpoint files, tries to read a different file name to the one written in `writeResult`, so at a minimum we should change it so they read the right files.

PR #230, opened 2025-10-21
Remove use global from LastConditions.f90

mcbarton:


tomneep (approved):
I'm going to approve/merge this with the caveat that we are currently not ever testing `lastConditions` (related to #206 I think)

PR #231, opened 2025-10-21
Replace nblocks with size(block) where possible

### LINKED ISSUE #18: Replace `nblocks` everywhere
We should be able to use e.g. `size(block)` in its place

mcbarton:
Closes #18


tomneep (approved):
I'll just quickly double check that the output doesn't change

Issue #234, opened 2025-10-23
`PressureForcing1` and `pressureForcingField` very similar


tomneep:
I think we can probably write a core subroutine that can cover both of these cases.

`diff -Bw <(git show f889209a9a:src/Forcing.f90 | sed -n '13,98p;98q') <(git show f889209a9a:src/Forcing.f90 | sed -n '860,943p;943q')`

```diff
1c1
< SUBROUTINE pressureForcing1(blk)
---
> SUBROUTINE pressureForcingField(blk)
4c4,5
<       REAL (dp) :: n1, pos1_x, pos1_y, pos1_z, pt1, aval, bval, cval, p_pos1, sur2nodeDis, dpdn, &
---
>       REAL (dp) :: n1, pos1_x, pos1_y, pos1_z, pt1, &
>                    aval, bval, cval, p_pos1, sur2nodeDis, dpdn, &
7d7
<       dpdn = 0._dp
8a9
>       dpdn = 0._dp
10c11
<  !$acc private (n, n1, pos1_x, pos1_y, pos1_z, pt1, aval, bval, cval, p_pos1, sur2nodeDis, dpdn,                   &
---
>  !$acc private (n1, pos1_x, pos1_y, pos1_z, pt1, aval, bval, cval, p_pos1, sur2nodeDis, dpdn,                   &
13,14c14
<  !$acc default(present)  &
<  !$acc private(derivatives)
---
>  !$acc default(present) private(derivatives)
17d16
<             !dpdn = 0.
37,39c36,38
<          dpdn = -((ac_z + at_z)*blk%cosGamma(blk%nelp(n)) + (ac_y + at_y) * &
<                   blk%cosBeta(blk%nelp(n)))-blk%yddot * &
<                      blk%cosBeta(blk%nelp(n))
---
>          dpdn = -((ac_z + at_z)*blk%cosGamma(blk%nelp(n)) &
>                 + (ac_y + at_y)*blk%cosBeta(blk%nelp(n))) &
>                 -blk%yddot*blk%cosBeta(blk%nelp(n))
47c46
<          pt1 = 1.5_dp*dsqrt(blk%deltax(i)**2 + blk%deltay(j)**2 + blk%deltaz(k)**2) &
---
>          pt1 = 1.51_dp*dsqrt(blk%deltax(i)**2 + blk%deltay(j)**2 + blk%deltaz(k)**2)&
56c55
<          DO il = 1, blk%nx+1
---
[...]

PR #235, opened 2025-10-23
Make common function for pressure forcing.

tomneep:
Shared by pressureForcing1 and pressureforcingfield.

Part of #234 (need to address the velocity equivalents too).

Putting as draft until tested on Baskerville.

### DIFF
--- a/src/Forcing.f90
+++ b/src/Forcing.f90
@@ -1,5 +1,5 @@
 module biocfd_forcing
-  use, intrinsic :: iso_fortran_env, only: dp => real64
+  use, intrinsic :: iso_fortran_env, only: dp => real64, int64
   use biocfd_interpolation, only: linear_interpolation, bilinear_interpolation
   use biocfd_blocks,only : Blocks
   implicit none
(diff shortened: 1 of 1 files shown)


src/Forcing.f90
```diff
@@ -10,18 +10,26 @@ module biocfd_forcing
...
+      integer :: low, high
       INTEGER :: n, k, j, i, il, jl, kl, i_x1, i_y1, i_z1
       REAL (dp) :: n1, pos1_x, pos1_y, pos1_z, pt1, aval, bval, cval, p_pos1, sur2nodeDis, dpdn, &
                    dpdn_e, ac_y, ac_z, at_y, at_z
+      real(dp) :: scale_factor
       real(dp) :: derivatives(3)
       dpdn = 0._dp
 
+      ! I expect this isn't needed, but is for a match with the old code
+      scale_factor = merge(1.51_dp, 1.5_dp, is_field)
```
warrickball, line 26: I don't think I've ever seen the `merge` function before. :exploding_head:

tomneep:
While this would be worthwhile in the long term, I think this is very low down the priority list. Can be reopened in future if needed.

PR #243, opened 2025-11-03
Make findtscells run on GPU

tomneep:
This is one of the remaining subroutines that isn't fully running on the GPU (see screenshot). This should be faster than looping through every cell on the CPU.

image


tomneep:
For reference, the trace of this branch looks like this, with the rough area of work highlighted, so we save ~85ms per iteration. Note that the GPU isn't working very efficiently here so there is maybe some optimisation possible but probably other areas with lower hanging fruit.

I think this is ready for review now @mcbarton.

image

mcbarton (approved):

PR #244, opened 2025-11-04
Add stress calc

tomneep:
This adds back the Stess Calculation that we had previously removed due to it being commented out. At the same time, we use a function that we defined in `Forcing` to reduce some of the repetition.

There is probably more refactoring to do, but this should be a starting point.

Not tested performance on GPU yet, I suspect that it might struggle to inline `compute_value_and_derivatives` from another module but let's see.


app/main.F90
```diff
@@ -134,6 +136,9 @@ PROGRAM main
         do g=blk_start, size(block)
            CALL pressureForcing1(block(g))
         end do
+        do g=blk_start, size(block)
+           call stressCal1(block(g), g, mu_f, rho_f, totime)
+        end do
```
tomneep: We should check where exactly we want this to be called. It would also be useful to know whether we always want this information, because then I'd say it would make more sense to just add it to the HDF5 file?

src/StressCalc.f90
```diff
@@ -0,0 +1,261 @@
...
+29  FORMAT("liftcoff_",I4.4,".dat")
+    OPEN(999, file=filename1, position="Append", status="unknown")
+    WRITE(999,*) viscousLiftcoefficient,  pressureLiftcoefficient, totime
+    CLOSE(999)
+
+  end subroutine stressCal1
+
+!> Find the position in the array where the value is greater than
+!> element i but less than element i+1
+pure function find_index_in_array(value, array, start, end) result(index)
```
tomneep: This is a function I originally wrote for Forcing, but hasn't made its way into dev yet. It may end up being move there (potentially we could move the `stressCal1` subroutine in there too if needed for inlining.

src/StressCalc.f90
```diff
@@ -0,0 +1,261 @@
...
+
+       normdis = diagdis
+       pos1_x = blk%xcent(ielem) + normdis * normal(1)
+       pos1_y = blk%ycent(ielem) + normdis * normal(2)
+       pos1_z = blk%zcent(ielem) + normdis * normal(3)
+
+       !**************************velocity and pressure at the surface**********************
+       IF (blk%ibSurfID(ielem)==50) THEN
+          blk% thetaDot  =  0.
+          blk% thetaDDot =  0.
```
tomneep, line 88: I don't quite get these lines (the same question for Forcing). Do we really need to store the new values of `thetaDot` and `thetaDdot`? Are they used anywhere beyond this block of code?

src/StressCalc.f90
```diff
@@ -0,0 +1,261 @@
...
+
+       dwdn_e = dot_product(derivatives, normal)
+       ddn_s(3) = (2._dp/normdis)*(w_pos1 - wsurf) - dwdn_e
+
+       !***********************calculate area of the elements******************
+       alen = blk%element_length(ielem)
+       area = alen/2._dp
+       area_yz = 0.5_dp*abs(alen * normal(1))
+       area_xz = 0.5_dp*abs(alen * normal(2))
+       area_xy = 0.5_dp*abs(alen * normal(3))
```
tomneep, line 153: This can be a one-liner, but need to think of a good variable name

tomneep:
Probably ready for a look from @Arjun-Shergill

src/StressCalc.f90
```diff
@@ -0,0 +1,261 @@
...
+    WRITE(filename1,19) id
+19  FORMAT("dragcoff_",I4.4,".dat")
+    OPEN(899, file=filename1, position="Append", status="unknown")
+    WRITE(899,*) viscousDragcoefficient, pressureDragcoefficient, totime
+    CLOSE(899)
+    WRITE(filename1,29) id
+29  FORMAT("liftcoff_",I4.4,".dat")
+    OPEN(999, file=filename1, position="Append", status="unknown")
+    WRITE(999,*) viscousLiftcoefficient,  pressureLiftcoefficient, totime
+    CLOSE(999)
```
mcbarton: Maybe we make use of new unit hhere. Also we can incorporate the format lines into the lines which write to filename1

src/StressCalc.f90
```diff
@@ -0,0 +1,261 @@
...
+       area_Sx = area_Sx + area_xz
+       area_Sy = area_Sy + area_yz
+
+    END DO
+    !$acc end parallel loop
+
+    area_Sx= 0.5_dp * area_Sx
+    area_Sy= 0.5_dp * area_Sy
+    ! area_Sx= 1.!0.5 * area_Sx
+    ! area_Sy= 1.!0.5 * area_Sy
```
mcbarton: These lines can be removed

src/StressCalc.f90
```diff
@@ -0,0 +1,261 @@
...
+          wsurf    = blk%thetaDot*(blk%ycent(ielem) - blk%piv_y)  ! + ydot
+          ac_z      = -blk%thetaDot**2*(blk%zcent(ielem) -blk% piv_z)
+          ac_y      = -blk%thetaDot**2*(blk%ycent(ielem) - blk%piv_y)
+          at_z      =  blk%thetaDDot*(blk%ycent(ielem) -blk% piv_y)
+          at_y      = -blk%thetaDDot*(blk%zcent(ielem) - blk%piv_z)
+       END IF
+       !       usurf     =  0.
+       !       wsurf     = blk% thetaDot*(blk%ycent(ielem) -blk% piv_y)
+       !       vsurf     = -blk%thetaDot*(blk%zcent(ielem) -blk% piv_z)
+       !       dpdn      = -((ac_z + at_z)*blk%cosAlpha(ielem)  + (ac_y + at_y)*blk%cosBeta(ielem))
```
mcbarton: Don't need these comments.

src/StressCalc.f90
```diff
@@ -0,0 +1,261 @@
...
+          vsurf    = -blk%thetaDot*(blk%zcent(ielem) - blk%piv_z) + blk%ydot
+          wsurf    = blk%thetaDot*(blk%ycent(ielem) - blk%piv_y)
+          ac_z      = -blk%thetaDot**2*(blk%zcent(ielem) -blk% piv_z)
+          ac_y      = -blk%thetaDot**2*(blk%ycent(ielem) -blk% piv_y)
+          at_z      = blk% thetaDDot*(blk%ycent(ielem)-blk% piv_y)
+          at_y      = -blk%thetaDDot*(blk%zcent(ielem)-blk% piv_z)
+       ELSE IF (blk%ibSurfId(ielem)==52) THEN
+          blk% thetaDot  = blk% thetaDot2
+          blk% thetaDDot = blk% thetaDDot2
+          usurf = 0._dp +blk%xdot
```
mcbarton: Remove all 0.0_dp +

src/StressCalc.f90
tomneep: I actually have a question about this. Why multiply this by `0.5` and then include a factor of `2` where it is used?

src/StressCalc.f90
tomneep: See comment in PR description about writing this information to HDF5. I don't think it is particularly a good idea to write the data like this full stop.

src/StressCalc.f90
Arjun-Shergill: I think what’s happening here is that the code calculates the aerodynamic coefficients during each iteration for surface elements belonging to either the body or one of the two wings (`blk%ibSurfID(ielem) == 50, 51, 52`). Since an element can't be in multiple regions at the same time, the integration effectively computes the total lift and drag for one of the two wings

To represent the total aerodynamic forces for the full butterfly (assuming symmetrical wings), the coefficients are therefore multiplied by two.

For clarity, the reference area is halved earlier (lines 215–216) because only one wing is being considered. Lines 219–222 then apply the standard lift and drag coefficient equations, but because its now Area/2, the coefficients end up being doubled to account for the full butterfly. It might be clearer to keep the halving of the area as-is, rather than removing it and replacing the factor of two in lines 219–222 with a four — even though mathematically, the results would be equivalent.

src/StressCalc.f90
tomneep: Removed

src/StressCalc.f90
tomneep: Removed

src/StressCalc.f90
```diff
@@ -0,0 +1,222 @@
...
+   ! TODO: No need for these to be int64
+   integer(int64), intent(in) :: start, end
+
+   !> The resulting index
+   integer :: index
+
+   ! Internal counter
+   integer :: i
+   !$acc routine seq
+   do i=start, end
```
mcbarton: Although there are no protected keywords in Fortran, its best to avoid keywords which are commonly used, so I think we should have a change the variable name end to something else.

src/StressCalc.f90
```diff
@@ -0,0 +1,222 @@
...
+    ! 3 array in Block_t
+    real(dp) :: normal(3)
+
+    viscousDrag = 0.
+    pressureDrag = 0.
+    viscousLift = 0.
+    PressureLift = 0.
+    surf_area = 0.
+    area_Sx = 0.
+    area_Sy = 0
```
mcbarton:
```suggestion
    viscousDrag = 0.0_dp
    pressureDrag = 0.0_dp
    viscousLift = 0.0_dp
    PressureLift = 0.0_dp
    surf_area = 0.0_dp
    area_Sx = 0.0_dp
    area_Sy = 0.0_dp
```

src/StressCalc.f90
```diff
@@ -0,0 +1,222 @@
...
+    !$acc          ddn_s, &
+    !$acc          alen, area, area_xz, area_yz, area_xy, &
+    !$acc          shear_force, f_surf, ac_z, ac_y, at_y, at_z)  &
+    !$acc default(present)    &
+    !$acc firstprivate (rho_f, mu_f) &
+    !$acc private(derivatives, normal) &
+    !$acc reduction(+: pressureDrag, viscousDrag, viscousLift, pressureLift, area_Sx, area_Sy, surf_area)
+    !DO ielem = 1, blk%ibElemCnt
+    DO ielem = 1, blk%ibElems
+       !  IF((blk%zcent(ielem).ge.0.0).and.(blk%zcent(ielem).le.2.0)) THEN
```
mcbarton: Is this comment explaining something, or can it be removed.

src/StressCalc.f90
```diff
@@ -0,0 +1,220 @@
+module biocfd_stress_calculation
+  use, intrinsic :: iso_fortran_env, only: dp => real64, int64
+  use biocfd_forcing, only: compute_value_and_derivatives
+  use biocfd_block_type, only: Block_t
+  IMPLICIT NONE
+
+  private
+
+  public :: stressCal1
```
mcbarton: Although there is almost certainly a stresscal2 defined somewhere, it might be worth renaming this to just stresscal, and if another stresscal subroutine gets integrated in the code at some point, then they can name it in a better way then just having a different number at the end. This is only a suggestion though, and am happy for you to keep this old naming convention.

src/StressCalc.f90
tomneep: This has all been removed now

src/StressCalc.f90
tomneep: Renamed to `stress_calculation`

mcbarton (approved):
Although there might be possibility for further improvement, this feels good to be merged in, and further improvements can be made in future PRs.

PR #245, opened 2025-11-04
Tagging th move blk argument

tomneep:
Adds the `blk` (and `id`) argument to `tagging_th_move`. I think this is the first one where we check that the interface corresponds to this block in this way but it will be a somewhat common practice once we get everything implemented for MPI.

For #165

### DIFF
--- a/src/Search.F90
+++ b/src/Search.F90
@@ -418,20 +418,16 @@ SUBROUTINE tagging_th_core(blk)
         INTEGER(int64), intent(in) :: id
         integer(int64) :: g
 
-        if (blk%move_check == 1) then
-           call tagging_th_core(blk)
-        endif
+        if (blk%move_check /= 1) return
+        call tagging_th_core(blk)
 
         DO g=1, size(intfr)
-           ! Find the interface which this block corresponds to. We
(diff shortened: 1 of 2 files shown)


src/Search.F90
```diff
@@ -412,31 +412,32 @@ SUBROUTINE tagging_th(blk,blk_no)
...
+        type(Block_t), intent(inout) :: blk
+        INTEGER(int64), intent(in) :: id
+        integer(int64) :: g
 
-        DO g=blk_start,size(block)
-        if ( block(g)%move_check == 1)then
-           call tagging_th_core(block(g))
+        if (blk%move_check == 1) then
+           call tagging_th_core(blk)
         endif
```
tomneep: I've trying to figure out if we could just do an early return here if `move_check /= 1`

mcbarton: tagging_th_core, fineUpdate_mv and fineUpdate_bd_mv will not be called if `move_check /= 1`, so you can do an early return. You should keep the  `if (intfr(g)%b_blk /= id) cycle` in the do loop though

tomneep: Done

mcbarton (approved):

PR #251, opened 2025-11-10
Add tests for hdf5 module

### LINKED ISSUE #237: HDF5 some variables not correctly set?
It looks like certain variables aren't being filled for some reason e.g.

```console
$ h5dump -d "block_00002/totime" output_00001.h5
HDF5 "output_00001.h5" {
DATASET "block_00002/totime" {
   DATATYPE  H5T_IEEE_F64LE
   DATASPACE  SCALAR
   DATA {
   (0): 0
   }
}
}
```

(I don't think it should be 0). Same for `zone[ijk]`.

mcbarton:
This PR

Fixes https://github.com/BioFSILab/Bio-CFD/issues/237
Fixes a previously bug we were unaware of where an attempt to create a hdf5 file with the integer writing hdf5 subroutine would cause errors (see h5f_acc_rdwr_f change to h5f_acc_excl_f)
Fixes Fortitudes preview rules for the hdf5 module (previously this PR https://github.com/BioFSILab/Bio-CFD/pull/250). Changes included by accident, but didn't see them worth reverting and including in a separate PR once realised.
Adds some tests for the hdf5 module (checks the ability to write scalar, 1D/2D/3D arrays for both real and integer types).

### DIFF
--- a/src/hdf5_io.F90
+++ b/src/hdf5_io.F90
@@ -3,11 +3,12 @@
 ! effectively empty
 #if USE_HDF5 == 1
 module biocfd_hdf5_io
-  use, intrinsic :: iso_fortran_env, only: dp => real64, int8,int16,int32,int64
+  use, intrinsic :: iso_fortran_env, only: dp => real64, int8,int32,int64
   use hdf5, only: hsize_t, hid_t, h5open_f, h5fopen_f, h5fcreate_f, h5lexists_f, &
        h5gcreate_f, h5gopen_f, h5screate_f, h5screate_simple_f, h5dcreate_f, &
        h5dwrite_f, h5dclose_f, h5sclose_f, h5gclose_f, h5fclose_f, h5close_f, &
-       h5F_acc_rdwr_f, H5S_SCALAR_F, H5T_STD_I64LE, h5f_acc_excl_f, h5t_native_double
+       h5F_acc_rdwr_f, H5S_SCALAR_F, H5T_STD_I64LE, h5f_acc_excl_f, h5t_native_double, &
+       h5acreate_f, h5awrite_f, size_t, h5aclose_f, h5sclose_f
   implicit none
 
   private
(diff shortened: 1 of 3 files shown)


src/hdf5_io.F90
```diff
@@ -27,18 +28,21 @@ subroutine hdf5_write_real(filename,scalar_input,&
     character(len=20) :: dataset_location
     integer(hsize_t),allocatable :: data_dims(:)
     integer(hid_t) :: file_id, dspace_id, dset_id, group_id
+    integer(hid_t) :: attr_id, aspace_id
+    integer(size_t), dimension(1) :: adims
```
tomneep:
```suggestion
    integer(size_t) :: adims(1)
```

src/hdf5_io.F90
```diff
@@ -152,18 +171,21 @@ subroutine hdf5_write_int(filename,scalar_input,&
...
```
tomneep:
```suggestion
    integer(size_t) :: adims(1)
```

tomneep (approved):
A couple of extremely minor suggestions that don't prevent approval. `dimension` seems a bit unnecessary for a single variable and so the suggestions make it more consistent with e.g. `integer(hsize_t),allocatable :: data_dims(:)`. Feel free to merge directly if you don't want to make the change.

Btw, did you test we can read the attributes into paraview?

mcbarton:
> A couple of extremely minor suggestions that don't prevent approval. `dimension` seems a bit unnecessary for a single variable and so the suggestions make it more consistent with e.g. `integer(hsize_t),allocatable :: data_dims(:)`. Feel free to merge directly if you don't want to make the change.
> 
> Btw, did you test we can read the attributes into paraview?

I have not tested I can read the attribute into Paraview. I suspect the python script will need updating to deal with the attribute case. I will investigate this, and revisit this issue in another PR if a change here is needed to get Paraview to be able to read in the attribute.

PR #253, opened 2025-11-12
First commit of MPI helpers

tomneep:
This adds a new module `biocfd_mpi_helpers` that will setup the strategy for splitting blocks across nodes. Currently, the strategy is taken from the MPI R&D branch (https://github.com/BioFSILab/Bio-CFD/blob/4d2640beabd0b865597af662eccd512c9fd6862c/app/main.F90#L54-L75) which is a pretty naive way of doing it as each node handles the same number of blocks, which won't be a good choice in the case you have 4 GPUs on node *A* and 1 GPU on node *B*.

I **think** that this is still a good starting point, and even if we change the algorithm the structure works the same. The idea is that `biocfd_init` gives us the loop arguments(?) that we'll use in main.

As a concrete example with the current implementation for two ranks, and ten blocks we'll get
```
Rank 0: start = 1, finish = 10, step=2
Rank 1: start = 2, finish = 10, step=2
```
The loop would be something like
```f90
do g=start, finish, step
  print *, g
end do
```
So rank 0: would print `1, 3, 5, 7, 9` and rank 1 would print `2, 4, 6, 8, 10`

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -35,9 +35,10 @@ PROGRAM main
         INTEGER (int64)   :: itamax, pcItaMax
         real(dp) :: aoa,aoa1,aoa2,phase_angle,piv_pt
         integer :: start, finish, step, rank
-        call biocfd_init(size(block), start, finish, step, rank)
 
         CALL readInput(surGeoPoints,char_f,istart,itamax,pcItaMax,aoa,phase_angle,piv_pt)
+        ! Need to init after we know the size of block
+        call biocfd_init(size(block), start, finish, step, rank)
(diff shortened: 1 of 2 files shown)


tomneep:
I'm going to mark this as ready for review. The `start`, `finish`, `step`, and `rank` variables aren't actually being used in `main` yet so this doesn't change any behaviour.

tomneep:
This is now ready to be re-reviewed @mcbarton

tomneep:
Tested with `nvfortran` on Baskerville

mcbarton (approved):

PR #255, opened 2025-11-14
Start to group together loops in main.f90

mcbarton:
Assuming I haven't missed anything when inspecting the subroutines, this PR groups together some of the loops in main without effecting the output (since each group of subroutine calls only have one subroutine which has print statements in it). I can almost certainty group together more, but I need to remove some of the random print statements in main.

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -1,5 +1,4 @@
-
-      PROGRAM main
+PROGRAM main
         use, intrinsic :: iso_fortran_env, only: int64, dp => real64
         USE global, only: block, blk_start, coarse_flcnt_check, deltat, &
              ita, ita1, totaltime, totime, &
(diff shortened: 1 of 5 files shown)


app/main.F90
tomneep: Can you delete all these lines too?

mcbarton: Done, and I used that commit to remove some of the random print statements in main, so that I can start grouping together more do loops in main.

app/main.F90
```diff
@@ -61,17 +55,12 @@ PROGRAM main
         do g=blk_start, size(block)
            CALL tagging_th(block(g),g)
         end do
-        print*,'11'
-        print*, "cellCount started"
         do g=blk_start, size(block)
            CALL cellCount_solid(block(g))
            print*,g, block(g)%fluidCellCount, block(g)%redCellCount, block(g)%blackCellCount
```
mcbarton: @tomneep do you find the final 2 print statements in main useful at all, or are you happy for me to remove them?

tomneep: Yes it is very useful for ensuring we aren't breaking anything. I think you can keep that and put "cellCount:" at the start of the line

```suggestion
           print*, "cellCount:", g, block(g)%fluidCellCount, block(g)%redCellCount, block(g)%blackCellCount
```

tomneep: Actually, since we are changing the output anyway, it might be best to put this line (still with the suggestion) into `cellCount_solid`

src/Blocks.f90
```diff
@@ -72,4 +72,3 @@ MODULE biocfd_block_type
     end type Block_t
 
 END MODULE biocfd_block_type
-
```
tomneep, line 75: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

mcbarton: Should I go through every .f90 and .F90 and just add the blank link new line at the end as part of the PR?

tomneep: No definitely not 🙂, but I wouldn't remove them if they're there

tomneep: It might be that these are right, and GitHub isn't showing them very well. All that matters is that the last line e.g. `END MODULE biocfd_block_type` ends in `\n`

tomneep (approved):

tomneep:
Tested with `nvfortran` on Baskerville successfully

PR #257, opened 2025-11-16
Do one hdf5 file per block instead of per iteration

mcbarton:
Instead of making a new hdf5 file per iteration, this makes it so that each block has a hdf5 file, which I believe is better suited to the case when we starting making mpi additions to the hdf5 file writing subroutine. This change however will break our python script which allows us to view the hdf5 file in Paraview.

### DIFF
--- a/src/WriteOutput_corner1.F90
+++ b/src/WriteOutput_corner1.F90
@@ -26,11 +26,11 @@ contains
        INTEGER  :: k, i, j
        REAL (dp), allocatable :: u1(:,:,:), v1(:,:,:), w1(:,:,:)
        character (len=11) :: dummy_1
-       character (len=5) ::dummy_2
+       character (len=14) ::dummy_2
 
          if (mod(ita,200_int64) /=0 .and. ita > 2) return
             write(dummy_1,'(A6,I5.5)') 'block_',blk_no
-            write(dummy_2,'(I5.5)') ita
+            write(dummy_2,'(A9,I5.5)') 'timestep_', ita
             allocate(u1(2:blk%nx+1,2:blk%ny+1,2:blk%nz+1),&
                      v1(2:blk%nx+1,2:blk%ny+1,2:blk%nz+1),&
                      w1(2:blk%nx+1,2:blk%ny+1,2:blk%nz+1))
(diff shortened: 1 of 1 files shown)


mcbarton:
Although this PR is ready for review, the python script for viewing the hdf5 file on Baskerville would need updating, and the changes need discussing with Chandans team before this would go in.

mcbarton:
This PR will be changed so we have a hdf5 file instead of per interation or per block to be a hdf5 for every block and iteration combination.

mcbarton:
@tomneep This PR is ready for review.

tomneep (approved):
This looks good to me. I probably wouldn't merge until we've figured out how to update the xdmf script

mcbarton:
> This looks good to me. I probably wouldn't merge until we've figured out how to update the xdmf script

Hi @tomneep I updated the xdmf script now if you want to try it out. To get it to work I executed the following from the Bio-CFD folder

```
python hdf5-paraview.py ./out/timestep_*_block_*.h5 >> output.xdmf
```

tomneep:
I know I previously approved this but I have just one question which I'll add now, and then I'm keen to get this merged!

src/WriteOutput_corner1.F90
```diff
@@ -26,11 +26,11 @@ SUBROUTINE write_output_hdf5(blk,blk_no)
        INTEGER  :: k, i, j
        REAL (dp), allocatable :: u1(:,:,:), v1(:,:,:), w1(:,:,:)
        character (len=11) :: dummy_1
-       character (len=5) :: dummy_2
+       character (len=15) :: dummy_2
 
          if (mod(ita,200_int64) /=0 .and. ita > 2) return
             write(dummy_1,"(A6,I5.5)") "block_",blk_no
-            write(dummy_2,"(I5.5)") ita
+            write(dummy_2,"(A9,I5.5,A1)") "timestep_", ita,"_"
```
tomneep: Here would it not be simpler to make have a single format string that includes `'block_` and `timestep_`? E.g. wouldn't 

```f90
write(dummy_2, "('timestep_', i5.5, '_block_', i5.5)") ita, blk_no
```
be clearer (ideally with a better variable name than `dummy`)?

tomneep: You could even just write the whole file path here right?

```
write(dummy_2, "('out/timestep_', i5.5, '_block_', i5.5, '.h5')") ita, blk_no
```

mcbarton: Applied suggestion and tested. We now no longer have a dummy_1 and dummy_2 variables. I also renamed filename1 to just filename.

Issue #209, opened 2025-10-02
MPI routines


tomneep:
Now that #165 is getting close to completion, we can start thinking about adding the actual MPI functions. The following functions and subroutines call `MPI` subroutines/functions in the MPI R&D branch.

## src/CoarseInterp.f90
- [x] coarseUpdate_newv #249 
- [x] coarseUpdate_pc #249 
- [x] coarseUpdate #249
## src/PcorVcor.F90
- [x] poissonSolver - #260 
## src/FineInterp_bound.f90
-  ~fineUpdate_bd~ (no longer exists)
- [x] fineUpdate_pc_bd #252
- [x] fineUpdate_newv_bd #252
- [x] fineUpdate_bd_mv - Called in `tagging_th_move` MPI routines moved to main #261  and `poissonSolver` #260 
## src/FineInterp.f90
- [x] fineUpdate_mv - Called once in `tagging_th_move` MPI routines moved to main #261 
## src/Search.F90
 - [x] block_move_check (we should probably move this into main) - Moved to main in #261 
## app/main.F90
 - [x] main #261 

So, actually, once things have been tidied a bit I think it will just be in the Interpolation modules that we need MPI communication (which makes sense) and maybe a few calls in main (which perhaps we can tidy away into a helper module). Things will get a bit more complicated if there can be interpolation between two different fine blocks, at the moment the assumption has been that there is only interpolation between fine blocks and the coarse block.

tomneep:
All of these should be addressed in one way another once #261 is merged

PR #260, opened 2025-11-17
Prepare PcorVcor for MPI

tomneep:
- Uses `biocfd_mpi_helpers` to control looping over blocks
- Adds `MPI_Bcast` before `fineUpdate_bd_mv` interface loop

Addresses `poissonSolver` task of #209

### DIFF
--- a/src/PcorVcor.F90
+++ b/src/PcorVcor.F90
@@ -122,7 +122,6 @@ module biocfd_pcor_vcor
 
         !$omp single
         CALL coarseUpdate_pc
-        g=1
         ! This is a slightly confusing loop but is written this way for MPI
         do g=start, finish, step
           if (g == 1) CALL REDBLACKSOR_linear(g, pcItaMax)
--- a/src/PcorVcor.F90
+++ b/src/PcorVcor.F90
@@ -122,7 +122,11 @@ module biocfd_pcor_vcor
 
         !$omp single
         CALL coarseUpdate_pc
-        ! This is a slightly confusing loop but is written this way for MPI
+        ! This is a slightly confusing loop but is written this way
+        ! for MPI. The reason it is like this is that when we run with
+        ! MPI we don't know where block(1) is. I'm not 100% that the
+        ! code will work if block(1) is anywhere other than rank 0,
+        ! but we ideally we shouldn't assume that it is.
         do g=start, finish, step
           if (g == 1) CALL REDBLACKSOR_linear(g, pcItaMax)
         end do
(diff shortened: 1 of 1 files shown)


tomneep:
Tested and works on Baskerville with `NVFORTRAN`

src/PcorVcor.F90
```diff
@@ -166,6 +188,8 @@ SUBROUTINE poissonSolver(pcItaMax)
               totalTime=totime + totalTime
           end do
 
+#ifndef BIOCFD_MPI
+          ! TN: Not going to do this if we are using MPI - I'm not sure how useful it is anyway
```
mcbarton, line 195: If this file is not useful, maybe we should just consider removing it, rather than keeping it for the non mpi build?

src/PcorVcor.F90
```diff
@@ -96,39 +113,44 @@ SUBROUTINE poissonSolver(pcItaMax)
...
         !$omp end do
 
         !$omp single
         CALL coarseUpdate_pc
         g=1
-        CALL REDBLACKSOR_linear(g,pcItaMax)
+        ! This is a slightly confusing loop but is written this way for MPI
+        do g=start, finish, step
+          if (g == 1) CALL REDBLACKSOR_linear(g, pcItaMax)
+        end do
```
mcbarton: Can you explain why it needs to be this way for mpi? I can't work out why.

If it needs to be this way, then surely the g=1 statement before the loop is irrelavent.

src/PcorVcor.F90
tomneep: Yes but maybe in a different PR?

src/PcorVcor.F90
tomneep: You are right that the `g=1` can be removed. I'll do that now.

src/PcorVcor.F90
```diff
@@ -96,39 +113,44 @@ SUBROUTINE poissonSolver(pcItaMax)
...
            ! Do not compute Red/Black here for block 1
-           if (g /= 1)  CALL REDBLACKSOR_linear(g,pcItaMax)
-
+           if (g /= 1)  CALL REDBLACKSOR_linear(g, pcItaMax)
         end do
         !$omp end do
 
         !$omp single
         CALL coarseUpdate_pc
         g=1
```
tomneep:
```suggestion
```

src/PcorVcor.F90
tomneep: The loop is needed because we don't know if the rank will have `block(1)` or not. There are other ways we could check, but this seemed fairly clean to me.

mcbarton: Makes sense to me now. Maybe just add a comment as in the code as to the loop being there because we don't know if the rank has block 1 or not.

If you don't think the comment is necessary, then feel free to resolve this comment and I will approve the PR.

tomneep: I think it is a good idea to add a more descriptive comment. I'm just testing something else at the moment but once that is finished I'll add the comment.

mcbarton (approved):

PR #265, opened 2025-11-27
Add gpu_helpers module (and use it)

tomneep:
This adds a new module `biocfd_gpu_helpers` that just has a single subroutine `set_gpu`. All it does is work out how many GPUs we have access to, what OMP thread we are on, and then assigns each thread to a GPU.

This is moving code out of `PcorVcor` since we'll use it in main too.

Suggestions for file/module names are welcome.

Setting to draft as needs testing in a multi-GPU environment to make sure it's actually doing something.


tomneep:
Hi @mcbarton. since we tested this last week I think it should be fine to be reviewed

mcbarton (approved):

PR #266, opened 2025-11-27
More parallel in main

tomneep:
Making this so I don't forget where I'm up to. Relies on #265 being merged first.

Note: There is an OpenMP region in `tagging_th` if OpenACC isn't used, which creates nested OMP parallel regions. By default (at least on my laptop), the inner OpenMP directives will just be ignored, unless nested OMP is explicitly enabled. This can be done with environment variables e.g.

`OMP_MAX_ACTIVE_LEVELS=2 OMP_NUM_THREADS=2,4`

will allow 2 levels, with the first level using 2 threads and the inner level using 4 threads. 

I *think* that a single value `OMP_NUM_THREADS` doesn't really work here because each level will use that many threads (so if we set `OMP_MAX_ACTIVE_LEVELS=2 OMP_NUM_THREADS=8` then we'll get 8 threads on level one and each of those 8 threads will spawn 8 threads. I'm not 100% sure of this however.

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -26,6 +26,7 @@ PROGRAM main
         use biocfd_forcing, only: pressureForcing1, pressureforcingfield, pressureforcingghost, &
              velocityforcing1, velocityforcingfield, velocityforcingghost
         use biocfd_mpi_helpers, only: biocfd_init, biocfd_finalize
+        use biocfd_gpu_helpers, only: set_gpu
 #ifdef BIOCFD_MPI
         use mpi_f08, only: MPI_Allreduce, MPI_Bcast, MPI_COMM_WORLD, MPI_DOUBLE_PRECISION, &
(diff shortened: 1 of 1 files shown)


tomneep:
Now that #265 is merged this can be reviewed too @mcbarton

mcbarton:
@tomneep can you fix the ci failure? The gfortran job has the following error

```
<ERROR> stopping due to failed compilation

app/main.F90:142:21:

  142 |            !$acc wait
      |                     1
Error: The !$ACC WAIT directive cannot be specified within a !$OMP PARALLEL region at (1)
STOP 1
```

tomneep:
Ah it is good that gcc spotted that because `nvfortran` doesn't seem to care! I'm 99% it can just be safely removed.

mcbarton (approved):

PR #268, opened 2025-12-02
Mpi helpers gpu aware

tomneep:
Tries to split work across nodes based on the GPUs available. 

For example, say we have 5 GPUs, with 4 on rank 0 and 1 on rank 1, and 10 blocks. This should work out that we have 5 GPUs in total, meaning we need 2 blocks per GPU. We should end up with 8 blocks on rank 0 and 2 blocks on rank 1.

~It should also deal with cases where blocks / GPUs isn't a round number. E.g. 11 blocks instead of 10 would lead to 2.2 blocks per GPU -> 8.8 blocks on rank 0 (gets rounded to 9) and 2.2 blocks on rank 1 (gets rounded to 2).~

~If we get a situation where there are multiple `x.5` values, then we trim off the last rank e.g. 10 blocks, 4 GPUs across 3 ranks (2, 1, 1), then we'd get 2.5 blocks per GPU, which would result in (5, 3, 3) blocks on ranks (0, 1, 2). In that case the upper bound is clamped such that we end up with (5, 3, 2).~

Cases where blocks / GPUs isn't a round number are dealt with using a "largest-remainder method".

There are probably more situations I've not considered.

### DIFF
--- a/src/PcorVcor.F90
+++ b/src/PcorVcor.F90
@@ -10,7 +10,6 @@ module biocfd_pcor_vcor
   use biocfd_fine_interp_bound, only : fineUpdate_newv_bd, fineUpdate_pc_bd, fineupdate_bd_mv
   use biocfd_coarse_update, only : coarseUpdate_newv, coarseUpdate_pc, coarseUpdate
   use biocfd_boundary_conditions, only : velocityBC
-  use biocfd_mpi_helpers, only: get_block_iteration_params
 #ifdef BIOCFD_MPI
    use mpi_f08, only: MPI_Bcast, MPI_COMM_WORLD, MPI_DOUBLE_PRECISION
 #endif
@@ -49,8 +48,6 @@ module biocfd_pcor_vcor
 
           ! Dummy values for omp/acc variables
           omp_threads = 1
-
-          call get_block_iteration_params(size(block), start, finish, step, rank)
 
 #ifdef _OPENMP
           ! Count the number of blocks on this rank (or in total if not
--- a/src/PcorVcor.F90
+++ b/src/PcorVcor.F90
@@ -21,7 +21,7 @@ module biocfd_pcor_vcor
 
   contains
 
-      SUBROUTINE poissonSolver(pcItaMax)
+      SUBROUTINE poissonSolver(pcItaMax, start, finish, step)
 
         INTEGER(int64) :: i, j,k, n, g
         INTEGER (int64),INTENT(IN)   :: pcItaMax
@@ -31,11 +31,11 @@ module biocfd_pcor_vcor
 #ifndef BIOCFD_MPI
         CHARACTER(len=160) :: filename1
 #endif
+        !> Variables to control iteration over blocks (mainly for MPI)
+        integer, intent(in) :: start, finish, step
 
         ! For controlling OpenMP
         integer :: omp_threads
-        ! Variables to control iteration over blocks (mainly for MPI)
-        integer :: start, finish, step, rank
 
           max_derrStdst=0._dp
           max_derr1=0._dp
--- a/src/mpi_helpers.f90
+++ b/src/mpi_helpers.f90
@@ -8,7 +8,7 @@ module biocfd_mpi_helpers
                      MPI_Finalize, MPI_Init_Thread, MPI_IN_PLACE, MPI_INTEGER, MPI_THREAD_SERIALIZED
 #endif
 #ifdef _OPENACC
-  use openacc, only: acc_device_default, acc_device_host, acc_get_num_devices, acc_set_device_type
+  use openacc, only: acc_device_default, acc_get_num_devices
 #endif
 
   implicit none
@@ -139,17 +139,10 @@ subroutine get_block_iteration_params_gpu(nblocks, start, finish, step, rank)
 
     rank_gpus(rank + 1) = acc_get_num_devices(acc_device_default)
 
-    ! The device type is set to the host due a crash that has been
-    ! seen in the AllGather operation. It seems that CUDA-aware MPI
-    ! can start doing things on the GPU that we don't really want it
-    ! to. Setting the device type to host here (and then back to the
-    ! GPU once the AllGather is complete) seems to not crash.
-    call acc_set_device_type(acc_device_host)
     ! Gather all GPUs together such that every rank has an array of
     ! GPU numbers
     call MPI_Allgather(MPI_IN_PLACE, 0, MPI_DATATYPE_NULL, rank_gpus, 1, MPI_INTEGER, &
                        MPI_COMM_WORLD, ierror)
-    call acc_set_device_type(acc_device_default)
     total_gpus = sum(rank_gpus)
 
     ! This is how many blocks we need to have per GPU
--- a/src/mpi_helpers.f90
+++ b/src/mpi_helpers.f90
@@ -133,6 +133,9 @@ subroutine get_block_iteration_params_gpu(nblocks, start, finish, step, rank)
     allocate(rank_gpus(world_size))
     ! This is the number of blocks we are going to place on each rank
     allocate(rank_blocks(world_size))
+    ! This is the "ideal" number of GPUs on each rank i.e. if we could
+    ! split blocks fractionally
+    allocate(ideal_blocks(world_size))
 
     rank_gpus(rank + 1) = acc_get_num_devices(acc_device_default)
 
--- a/src/mpi_helpers.f90
+++ b/src/mpi_helpers.f90
@@ -141,7 +141,7 @@ subroutine get_block_iteration_params_gpu(nblocks, start, finish, step, rank)
     rank_blocks = nint(rank_gpus * real_blocks_per_gpu)
 
     start = sum(rank_blocks(1:rank))
-    finsh = sum(rank_blocks(1:rank+1))
+    finish = sum(rank_blocks(1:rank+1))
     step = 1
     print *, "Testing - start = ", start, "finish = ", finish
 #else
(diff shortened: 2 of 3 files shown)


src/PcorVcor.F90
```diff
@@ -50,8 +49,6 @@ SUBROUTINE poissonSolver(pcItaMax)
           ! Dummy values for omp/acc variables
           omp_threads = 1
 
-          call get_block_iteration_params(size(block), start, finish, step, rank)
```
tomneep, line 53: It didn't really make sense to call this here because it involved `MPI` call every iteration. Instead I added `start`, `finish`, and step` as arguments to the subroutine

src/mpi_helpers.f90
```diff
@@ -91,4 +94,74 @@ subroutine get_block_iteration_params(nblocks, start, finish, step, rank)
 
 end subroutine get_block_iteration_params
 
+!> Work out how work will be shared across MPI nodes when we have
+!> multiple GPUs! The idea here is that we will work out how many GPUs
+!> our rank has and how many GPUs there are overall. We can then split
+!> the blocks proportionally across nodes. Each rank must have at
+!> least one GPU for this to work.
+subroutine get_block_iteration_params_gpu(nblocks, start, finish, step, rank)
```
tomneep: Once we've given this a review we can possibly merge this with `get_block_iteration_params_gpu` so we have one big block of 

```F90
#if defined(BIOCFD_MPI) && defined(_OPENACC)
! MPI + GPU case
...
#elif defined(BIOCFD_MPI)
! MPI (no GPU)
...
#else
! No MPI
```

src/mpi_helpers.f90
```diff
@@ -91,4 +94,72 @@ subroutine get_block_iteration_params(nblocks, start, finish, step, rank)
...
+    finish = sum(rank_blocks(1:rank+1))
+
+    ! Due to the nint, it may be that we end up with sum(rank_blocks)
+    ! greater than nblocks. In this case, we'll just cap the finish
+    ! value to nblocks. Need to think if there is a better way to deal
+    ! with this.
+    finish = min(finish, nblocks)
+
+    step = 1
+    print *, "MPI config - rank = ", rank, "gpus = ", rank_gpus, &
```
tomneep: This is pretty useful to have I think, but should be moved somewhere else so we get it regardless of MPI setup

tomneep:
This is a bit tricky so it would be good if both @mcbarton and @warrickball could have a look through this and see if it makes some sense (and think up tricky edge cases)

src/mpi_helpers.f90
tomneep: (and just to be clear this subroutine isn't currently used in this PR!)

warrickball:
Interesting little problem! I don't have a solution but I've got some fresh problems. Assuming I've copied your algorithm correctly to a crude little Python snippet, I have an issue sharing 9 blocks across nodes with `rank_gpus = [2,2,2,2]`. The bounds I get are `[0, 2, 4, 6, 8]`, which leaves one block unaccounted for. I think this is because `9/sum(rank_gpus)` is 1.125 and we always get rounded down.

I don't have an immediate idea of how best to solve this. I had a quick search for whether this is an example of a standard problem and I don't think any translate to this case.

One possibility might be to share the work by computing what each rank's (fractional) share of the workload should be, then loop over the number of blocks, incrementing the blocks assigned to which ever node is furthest from its "fair share".

FWIW, these are the two Python files I've been using to experiment (with apologies that these are _very_ hacky, but I won't have time for this tomorrow):

<details>
<summary> <code>share.py</code> </summary>

```py
import numpy as np

def fortran_round(x):
    """Reproduce Fortran rounding, which is always upwards for x.5,                                                                                                                                                                                           
    unlike Python's rounding to even."""
    return np.where(x%1.0 == 0.5, np.round(x+0.0001), np.round(x))

def nint(x):
    """Reproduce Fortran's ``nint`` function."""
    return fortran_round(x).astype(int)

def share(nblocks, rank_gpus):
    """Given ``nblocks`` to share across the GPUs per node defined in                                                                                                                                                                                         
    the list ``rank_gpus``, returns a list of the start and end                                                                                                                                                                                               
    indices of the blocks.  The output will have one more element than                                                                                                                                                                                        
    ``rank_gpus``.                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                              
    >>> print(share(4, [2, 2]))                                                                                                                                                                                                                               
    [0 2 4]                                                                                                                                                                                                                                                   
    >>> print(share(4, [1, 1, 1, 1]))                                                                                                                                                                                                                         
    [0 1 2 3 4]                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                              
    """
    total_gpus = sum(rank_gpus)
    real_blocks_per_gpu = nblocks/total_gpus

    rank_blocks = nint(np.array(rank_gpus)*real_blocks_per_gpu)
    return np.minimum(np.hstack([0, np.cumsum(rank_blocks)]), nblocks)

if __name__ == "__main__":
    print(share(4, [2,2]))
```

</details>

<details>
<summary> <code>test_share.py</code> </summary>

```py
import unittest
import numpy as np
from share import share, fortran_round

class TestShare(unittest.TestCase):
    def test_fortran_round(self):
	np.testing.assert_equal(fortran_round(np.array([1.5, 2.5])), [2, 3])

    def test_one_block_per_gpu(self):
        # each line is
        # np.testing.assert_equal(share(nblocks, [gpus-on-rank1, gpus-on-rank2, ec]), expected edges)
        np.testing.assert_equal(share(4, [2, 2]), [0, 2, 4])
	np.testing.assert_equal(share(10, [4, 1]), [0, 8, 10])
        np.testing.assert_equal(share(11, [4, 1]), [0, 9, 11])
        np.testing.assert_equal(share(10, [2, 1, 1]), [0, 5, 8, 10])
        np.testing.assert_equal(share(10, [4, 2, 3, 1]), [0, 4, 6, 9, 10])
        np.testing.assert_equal(share(9, [2, 2, 2, 2]), [2, 4, 6, 9])
        np.testing.assert_equal(share(11, [4, 2, 3, 1]), [0, 5, 7, 10, 11])
```

</details>

src/mpi_helpers.f90
```diff
@@ -91,4 +94,73 @@ subroutine get_block_iteration_params(nblocks, start, finish, step, rank)
...
+
+    ! TODO: More testing is needed
+    start = sum(rank_blocks(1:rank)) + 1
+    finish = sum(rank_blocks(1:rank+1))
+
+    ! Due to the nint, it may be that we end up with sum(rank_blocks)
+    ! greater than nblocks. In this case, we'll just cap the finish
+    ! value to nblocks. Need to think if there is a better way to deal
+    ! with this.
+    finish = min(finish, nblocks)
```
warrickball: If you move the `allocate` further up, you can possibly isolate the core of the algorithm so that you can test it in isolation. I'm personally finding it sufficiently messy that I think it might be worthwhile.

tomneep:
Thanks for having a look @warrickball. I've changed the algorithm to try and account for the case you found after a bit of reading. 

A slightly separate issue is that I'm getting occasional segfaults that appear to be caused by CUDA-aware MPI. This can be disabled by `export OMPI_MCA_opal_cuda_support=0` but I don't really understand why it is happening (and whether we are losing an performance by doing this).

warrickball:
I've just experimented a bit further with the new algorithm and I haven't managed to fault it.

tomneep:
Replacing `AllGather` with `Gather` then `Bcast` seems to fix the problems I was seeing, so I'll answer my own review comments now.

src/mpi_helpers.f90
tomneep: This is done now (and algorithm changed somewhat anyway)

src/mpi_helpers.f90
tomneep: I decided to leave this here for now, but only print the GPUs on this particular rank, which I think makes the information a bit easier to parse

mcbarton:
@tomneep Once you fix the trailing whitespace issue, I will approve and take this in.

tomneep:
Done, but it might be worthing getting another look from @warrickball

src/mpi_helpers.f90
```diff
@@ -63,15 +66,90 @@ subroutine biocfd_finalize()
...
+      j = maxloc(ideal_blocks - rank_blocks, dim=1)
+      rank_blocks(j) = rank_blocks(j) + 1
+    end do
+
+    start = sum(rank_blocks(1:rank)) + 1
+    finish = sum(rank_blocks(1:rank+1))
+    step = 1
+    print *, "MPI config - rank = ", rank, "gpus = ", this_rank_gpus, &
+             "start = ", start, "finish = ", finish
+#elif defined(BIOCFD_MPI)
```
warrickball, line 152: I hadn't looked at the code below because it's unchanged but it now looks to me (at a very hurried glance) like `step` isn't given a value in this block, perhaps because a previous `step = 1` has entered one of the other bits of the directive `#if/#elif/#else` block.

I'm not sure of the logic. If there are no GPUs, do we not want to send equal amounts of work to each MPI rank?

I'm also 100% happy for this to become part of an issue for later, since we aren't really worried about MPI+CPU (no GPU) usage.

warrickball:
I've had a hurried look and the core GPU-sharing scheme is unchanged and I think the changes to the subroutine calls makes sense, so happy for this to be merged.

I've left a comment about what happens in the MPI+CPU case (no GPUs) but that can become an issue for later.

src/mpi_helpers.f90
tomneep: I had noticed the same thing and it is a bit confusing, but `step` is set by `call MPI_Comm_size(MPI_COMM_WORLD, step, ierror)`. I wonder if we should add an explicit `step = step` or something?

tomneep: And also agree that this case isn't really something we should worry too much about. I suppose really if there are no GPUs we should split by cores but that seems like too much effort for something we aren't going to use

warrickball: Ah, so it does. No need to do anything here then.

warrickball (approved):

mcbarton (approved):

PR #269, opened 2025-12-05
Add Hdf5 reader subroutines

mcbarton:
This PR adds the ability to read 3d,2d,1d and scalar arrays from hdf5 files for both integer and real types.

### DIFF
--- a/fpm.toml
+++ b/fpm.toml
@@ -1,7 +1,7 @@
 name = "Bio-CFD"
 # To use HDF5 uncomment the dependencies.hdf5 line below and set
 # USE_HDF5=1
-# dependencies.hdf5 = "*"
+dependencies.hdf5 = "*"
 version = "0.1.0"
 [build]
 auto-executables = true
@@ -10,7 +10,7 @@ auto-examples = true
 external-modules = ["hdf5", "openacc", "mpi_f08"]
 
 [preprocess]
-cpp.macros = ["USE_HDF5=0"]
+cpp.macros = ["USE_HDF5=1"]
 
 
 [install]
--- a/fpm.toml
+++ b/fpm.toml
@@ -1,7 +1,7 @@
 name = "Bio-CFD"
 # To use HDF5 uncomment the dependencies.hdf5 line below and set
 # USE_HDF5=1
-dependencies.hdf5 = "*"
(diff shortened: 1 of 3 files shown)


tomneep:
Make sure you add a description to the PR. Can you remind me where we got to with assumed-rank arrays? Would that simplify things?

mcbarton:
> Make sure you add a description to the PR. Can you remind me where we got to with assumed-rank arrays? Would that simplify things?

Although assumed rank is a Fortran feature that would simplify the subroutines considerably, its not supported by the the nvfortran compiler.

tomneep:
Ah ok, I had misremembered that was one of the reasons we installed a newer version for assumed rank but it was probably for something else.

mcbarton:
> Ah ok, I had misremembered that was one of the reasons we installed a newer version for assumed rank but it was probably for something else.

I think it was the reason why we had it installed, since some post said that the newer version supports it, but it ended up not being true. I've seen posts since that says nvfortran will not be adding support for more Fortran features, and is waiting until Flang is more developed, as which point they will switch nvfortran to that.

tomneep:
That is disappointing. Since I don't really have any past HDF5 experience I can't really comment on it (@warrickball might be better placed), but this seems like an awful amount of boilerplate.

warrickball:
To be honest, I've never written Fortran HDF5 I/O code myself but the library I'm used to having under the hood is [ForUM](https://github.com/rhdtownsend/forum). I've just been browsing around the [relevant code there](https://github.com/rhdtownsend/forum/blob/main/src/lib/hdf5io_m.fypp) and it doesn't seem easy (or even possible) to avoid individual readers for each shape, at least not without going to an external preprocessor to generate code. We're handling HDF5 I/O internally and we're in control of the HDF5 data structures anyway, so I think this is fine.

ForUM uses the fypp preprocessor to expand a whole lot of private reader subroutines, then I think uses the dimension (and data type) of a pointer `target` to infer which of the subroutines to call from one public `read_dset` function. It makes the library API tidier but I don't think there's less code. 

I don't know if `nvfortran` would even build something like ForUM's interface.

tomneep (approved):
It sounds like this is the best we can do for now. Approving and merging.

PR #272, opened 2025-12-11
Restore checkpointing

### LINKED ISSUE #206: Checkpointing

mcbarton:
Closes https://github.com/BioFSILab/Bio-CFD/issues/206

Dependent on https://github.com/BioFSILab/Bio-CFD/pull/269

As an example of restarting, copy the hdf5 checkpointing files from out/Checkpoint

timestep_00004_block_00001.h5  timestep_00004_block_00002.h5  timestep_00004_block_00003.h5

to

input/Checkpointing/

Checkpoint_1.h5  Checkpoint_12.h5  Checkpoint_3.h5

and change istart to 1.

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -2,7 +2,7 @@ PROGRAM main
         use, intrinsic :: iso_fortran_env, only: int64, dp => real64
         USE global, only: block, coarse_flcnt_check, deltat, &
              ita, ita1, totaltime, totime, &
-             pi,uc,re,intfr
+             pi,uc,intfr
         use biocfd_search, only: findDistnode, shiftSurfaceNodesInitial, computeSurfaceNorm, &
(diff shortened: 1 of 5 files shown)


tomneep:
Has something strange happened with the commits here? We aren't expecting the changes in hdf5_io.F90 to show up in the diff are we?

mcbarton:
Replaced by https://github.com/BioFSILab/Bio-CFD/pull/286 which doesn't have the diff issues

PR #273, opened 2025-12-12
Make nvhpc ci job build with mpi

mcbarton:
This PR makes the nvhpc ci job build Bio-CFD with mpi. This is done by appending to FPM_FFLAGS the flags for the mpif90 compiler, and by adding BIOCFD_MPI to the proprocessor macros in the fpm.toml (done this way to avoid mpi builds being the default.

### DIFF
--- a/.github/workflows/setup-fortran.yml
+++ b/.github/workflows/setup-fortran.yml
@@ -67,6 +67,14 @@ jobs:
           # Install hdf5 dependency
           sudo apt install libhdf5-dev
 
+      - name: Modify FPM_FFLAGS to include mpi flags nvhpc
+        if: ${{ contains(matrix.toolchain.compiler, 'nvidia-hpc') }}
+        run: |
+          export MPI_FLAGS=$(mpif90 -show | cut -d' ' -f2-)
+          export FPM_FFLAGS="$MPI_FLAGS $FPM_FFLAGS"
+          echo "FPM_FFLAGS=$FPM_FFLAGS" >> $GITHUB_ENV
+          sed -i 's/"USE_HDF5=0"/"USE_HDF5=0","BIOCFD_MPI"/' fpm.toml
+
       - run: |
           # Build and test code using FPM
           fpm test


tomneep:
Nice- do you want to keep the none MPI version in the CI too? I know that will use more minutes...

mcbarton:
> Nice- do you want to keep the none MPI version in the CI too? I know that will use more minutes...

We aren't using any more minutes per ci run than we was previously with the way the PR is at the moment. The gfortran job has stayed as non mpi due to issues with the gfortran version. The NVHPC job has become the mpi job. 

The non mpi job is useful in that checks we don't introduce any new warnings.

tomneep:
Yes I understand we aren't using any more minutes as this PR is. My question was do you think it would be worthwhile to have nvfortran with and without MPI? I think there is some argument to keep a none MPI version of nvfortran, but it isn't essential. If we had unlimited minutes for example I'd definitely have nvfortran with and without MPI.

tomneep (approved):

.github/workflows/setup-fortran.yml
```diff
@@ -67,6 +67,14 @@ jobs:
           # Install hdf5 dependency
           sudo apt install libhdf5-dev
 
+      - name: Modify FPM_FFLAGS to include mpi flags nvhpc
+        if: ${{ contains(matrix.toolchain.compiler, 'nvidia-hpc') }}
+        run: |
+          export MPI_FLAGS=$(mpif90 -show | cut -d' ' -f2-)
+          export FPM_FFLAGS="$MPI_FLAGS $FPM_FFLAGS"
+          echo "FPM_FFLAGS=$FPM_FFLAGS" >> $GITHUB_ENV
+          sed -i 's/"USE_HDF5=0"/"USE_HDF5=0","BIOCFD_MPI"/' fpm.toml
```
warrickball, line 76: I'm sure I'm just missing something and am very late, but do we not want `USE_HDF5=1`, rather than `0`?

Issue #204, opened 2025-10-02
Better precision integers (and possibly reals)


tomneep:
There are a few cases in the code where we have integers which I suspect are much larger than they'd even need to be. For example, in main we define `integer (int64) :: g` as the block index. While we want to be ambitious, it seems unlikely we'll ever be running with over 2 billion butterflies!

PR #275, opened 2025-12-12
Forcing find index in array

tomneep:
This is just a "code tidy" PR so certainly not high priority, but I've started this several times and never got round to finishing it. 

Moves `find_index_in_array` from `StressCalc` to `Forcing` and is used. Can be used in all subroutines.

- [x] pressureForcing1
- [x] pressureForcingField
- [x] pressureForcingGhost
- [x] velocityForcing1
- [x] velocityForcingField
- [x] velocityForcingGhost

There are a bunch of `int64`s scattered around which are annoying, but that is part of #204.

### DIFF
--- a/src/StressCalc.f90
+++ b/src/StressCalc.f90
@@ -1,6 +1,6 @@
 module biocfd_stress_calculation
   use, intrinsic :: iso_fortran_env, only: dp => real64, int64
-  use biocfd_forcing, only: compute_value_and_derivatives
+  use biocfd_forcing, only: compute_value_and_derivatives, find_index_in_array
   use biocfd_block_type, only: Block_t
   IMPLICIT NONE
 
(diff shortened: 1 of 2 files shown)


src/StressCalc.f90
```diff
@@ -1,6 +1,6 @@
 module biocfd_stress_calculation
   use, intrinsic :: iso_fortran_env, only: dp => real64, int64
-  use biocfd_forcing, only: compute_value_and_derivatives
+  use biocfd_forcing, only: compute_value_and_derivatives, find_index_in_array
```
mcbarton, line 3: Has find_index_in_array been moved to Forcing.f90 just because its used more in that module, or for another reason? Feels like because it is generic enough subroutine, it could probably have its own module if being moved, and in a subsequent PR compute_value_and_derivatives could be moved to that new module.

mcbarton (approved):

PR #277, opened 2025-12-15
Blk 3 digit

tomneep:
Should be merged along with https://github.com/BioFSILab/biocfd-test-input/pull/6. 

The most important change is to read zero padded blocks (3 digits), but I also changed the file opening to use `newunit` added `status="old"` and `action="read"` and then indented the file reads to be a bit nicer.


tomneep:
I should add this is required for the upcoming 65 butterfly case (and should work for up to 999 butterflies)

mcbarton (approved):

PR #281, opened 2025-12-19
Add one more region of parallelism to PcorVcor

tomneep:
Ensures that zeroing of arrays on the GPU is done in parallel when there are multiple GPUs available.


tomneep:
Ok... pretty sure it works on Baskerville. I have an idea to fix it.

tomneep:
Pretty sure this should work now but will leave as draft for now as getting other things documented is more important. It is a performance optimisation but not a critical one.

PR #285, opened 2026-01-02
Move all remaining integer and real variables out of global module

### LINKED ISSUE #141: Stop having global variables
The eventual goal is to having it so Bio-CFD no longer has a global module, and there each subroutine has either local variables or arguments explicitly stating what it is changed in that subroutine.

mcbarton:
This PR moves all remaining integer and real variables out of global module

### DIFF
--- a/app/main.F90
+++ b/app/main.F90
@@ -32,18 +32,18 @@ PROGRAM main
         IMPLICIT NONE
 
         INTEGER (int64) :: g
-        INTEGER (int64)   :: surGeoPoints, ita, ita1, inor, blk_start
+        INTEGER (int64)   :: surGeoPoints, ita=0, ita1=0, inor
         CHARACTER (LEN = 3)   :: char_f
         INTEGER               :: istart
         INTEGER (int64)   :: itamax, pcItaMax, coarse_flcnt_check
         real(dp) :: aoa, aoa1, aoa2, phase_angle, piv_pt, mu_f, rho_f
-        real(dp) :: deltat, re, totalTime, totime, uc, u0, dt_order, dxmin, epsi, freq, &
+        real(dp) :: deltat, re, totalTime, totime=0.0_dp, uc, dxmin, epsi, freq, &
              omega1, omega2, omega3, omega4
         integer :: start, finish, step, rank
 
         CALL readInput(surGeoPoints, char_f, istart, itamax, pcItaMax, aoa, phase_angle, piv_pt, &
-                       mu_f, rho_f, ita, ita1, inor, blk_start,deltat,dt_order,dxmin, &
-                           epsi,freq,omega1,omega2,omega3,omega4, re, totime, u0, uc)
+                       mu_f, rho_f, inor, deltat,dxmin, &
+                           epsi,freq,omega1,omega2,omega3,omega4, re, uc)
         ! Need to init after we know the size of block
         call biocfd_init(size(block), start, finish, step, rank)
         CALL readBlockInterface
--- a/app/main.F90
+++ b/app/main.F90
@@ -1,7 +1,7 @@
 PROGRAM main
         use, intrinsic :: iso_fortran_env, only: int64, dp => real64
-        USE global, only: block, coarse_flcnt_check, deltat, &
-             ita, ita1, totaltime, totime, &
+        USE global, only: block, deltat, &
+              totaltime, totime, &
              pi,uc,re,intfr
         use biocfd_search, only: findDistnode, shiftSurfaceNodesInitial, computeSurfaceNorm, &
              tagging_th, tagging_th_move, block_move_check, cellcount_solid, &
@@ -34,15 +34,15 @@ PROGRAM main
         IMPLICIT NONE
 
         INTEGER (int64) :: g
-        INTEGER (int64)   :: surGeoPoints
+        INTEGER (int64)   :: surGeoPoints, ita, ita1, inor, blk_start
         CHARACTER (LEN = 3)   :: char_f
         INTEGER               :: istart
-        INTEGER (int64)   :: itamax, pcItaMax
+        INTEGER (int64)   :: itamax, pcItaMax, coarse_flcnt_check
         real(dp) :: aoa, aoa1, aoa2, phase_angle, piv_pt, mu_f, rho_f
         integer :: start, finish, step, rank
 
         CALL readInput(surGeoPoints, char_f, istart, itamax, pcItaMax, aoa, phase_angle, piv_pt, &
-                       mu_f, rho_f)
+                       mu_f, rho_f, ita, ita1, inor, blk_start)
         ! Need to init after we know the size of block
         call biocfd_init(size(block), start, finish, step, rank)
         CALL readBlockInterface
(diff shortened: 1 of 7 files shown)


mcbarton:
@tomneep this PR is ready for review.

tomneep (commented):
Hi @mcbarton, sorry I missed this. 

I know we have an issue about this but I'm not really sure that this really makes the code clearer or more maintainable? The fact that we end up with a couple of subroutines taking so many arguments is a bit concerning. It is probably too late now but perhaps a neater way of doing this would have been to have a `type` that holds some of these variables that we pass around. But then since there would just be a single instance of that type anyway you might argue it could just be a module. 

Any thoughts @warrickball?

warrickball:
I've started and deleted about three replies now because I think I'm basically undecided. I've certainly seen this pattern before, so I'm used to seeing such long procedure signatures in Fortran. I think it's easier to understand each procedure with explicit (even if verbose) signatures. I expect it'd make testing easier too. It's an easy gotcha for a researcher to start using a procedure without realising that it's actually using something from the `global` module.

I agree a `type` would make sense but I'm not sure we know enough about the domain to know which variables to group together (though maybe an `inputs` type or similar would be enough).

tomneep:
I don't doubt it crops up a lot in Fortran code (especially old code). Ok I'm happy for this to go in, but I'll make a couple of suggestions for places we might get away not passing arguments around.

src/ReadInput.f90
```diff
@@ -16,11 +12,14 @@ module biocfd_read_input
   contains
 
       SUBROUTINE readInput(surGeoPoints,char_f,istart,itamax,pcItaMax,aoa,phase_angle,piv_pt, &
-                           mu_f, rho_f)
+                           mu_f, rho_f, ita, ita1, inor, blk_start,deltat,dt_order,dxmin, &
+                           epsi,freq,omega1,omega2,omega3,omega4, re, totime, u0, uc)
        INTEGER (int64) :: i, g,io
        CHARACTER(len=160)  :: filename1
-       INTEGER (int64),INTENT(OUT)   :: surGeoPoints, itamax,pcItaMax
+       INTEGER (int64),INTENT(OUT)   :: surGeoPoints, itamax,pcItaMax, ita, ita1, inor, blk_start
```
tomneep: I think we can remove `ita`, `ita1`, and `totime` from `readInput`. The only thing that happens is they are set to 0, so we can move that into main and remove the pointless print.

tomneep: I think `blk_start` can be made a parameter of the `biocfd_read_input` module (it is set to 2 and never changed). It doesn't seem to be used anywhere else?

src/ReadInput.f90
```diff
@@ -16,11 +12,14 @@ module biocfd_read_input
...
        REAL(dp),intent(out) :: aoa,phase_angle,piv_pt, mu_f, rho_f
+       real(dp), intent(out) :: deltat, dt_order, dxmin, epsi, freq, omega1,omega2, omega3, &
```
tomneep: Is `dt_order` used anywhere beyond this subroutine?

tomneep: The `omega` parameters are a bit of a mess (I will open an issue for this and a PR) but they can stay in this PR.

src/ReadInput.f90
```diff
@@ -16,11 +12,14 @@ module biocfd_read_input
...
+             omega4,  re, totime, u0, uc
```
tomneep: Is `u0` used anywhere beyond this subroutine?

src/Search.F90
```diff
@@ -49,9 +46,11 @@ SUBROUTINE findDistnode(blk)
         END DO
         end subroutine findDistnode
 
-        SUBROUTINE shiftSurfaceNodesInitial(blk,aoa1,aoa2,piv_pt)
+        SUBROUTINE shiftSurfaceNodesInitial(blk,aoa1,aoa2,piv_pt, ita, &
+             deltat, dxmin, totime)
         type(Block_t), intent(inout) :: blk
-        real(dp),intent(in) :: aoa1, aoa2, piv_pt
+        real(dp),intent(in) :: aoa1, aoa2, piv_pt, deltat, dxmin, totime
```
tomneep: I realise I'm a bit confused about `aoa1` and `aoa2`. From what I can tell the values are redefined from scratch in `computeSurfaceVariables`. Can we just define the variables inside `shiftSurfaceNodesInitial` rather than in `main` (i.e. pass `aoa` instead of `aoa1` and `aoa2`)?

tomneep (commented):
I had a quick look at where we might be able to get rid of a few things. Worth double checking.

src/ReadInput.f90
tomneep: Opened issue #295 for the `omega` values

src/ReadInput.f90
```diff
@@ -39,12 +40,13 @@ SUBROUTINE readInput(surGeoPoints,char_f,istart,itamax,pcItaMax,aoa,phase_angle,
...
        REAL(dp) :: alpha_m,theta_m,alpha_m1,theta_m1, l_c,u_tip,disp
        INTEGER (int64) :: intflines, nblocks
+       real(dp) :: dt_order, u0
        NAMELIST /input_data/ nblocks, intflines,  &
                    itamax, epsi, pcItaMax,omega1,omega2,omega3,omega4, &
                    re,rho_f, mu_f, l_c, &
                    u0,  &
                    surGeoPoints, phase_angle, freq, aoa, piv_pt,alpha_m, theta_m, &
-                   istart, dt_order,  inor, dxmin
+                   istart,  inor, dxmin
```
tomneep: I think you meant to leave this in?

mcbarton: Yes, missed that. Fixed now.

app/main.F90
```diff
@@ -34,15 +32,19 @@ PROGRAM main
...
         INTEGER (int64) :: g
-        INTEGER (int64)   :: surGeoPoints
+        INTEGER (int64)   :: surGeoPoints, ita, ita1, inor
         CHARACTER (LEN = 3)   :: char_f
         INTEGER               :: istart
-        INTEGER (int64)   :: itamax, pcItaMax
-        real(dp) :: aoa, aoa1, aoa2, phase_angle, piv_pt, mu_f, rho_f
+        INTEGER (int64)   :: itamax, pcItaMax, coarse_flcnt_check
+        real(dp) :: aoa, phase_angle, piv_pt, mu_f, rho_f
+        real(dp) :: deltat, re, totalTime, totime=0.0_dp, uc, dxmin, epsi, freq, &
```
tomneep: I suppose we want this to be defined below for consistency with other definitions (although I agree this is probably fine)

src/Search.F90
```diff
@@ -49,12 +46,18 @@ SUBROUTINE findDistnode(blk)
         END DO
         end subroutine findDistnode
 
-        SUBROUTINE shiftSurfaceNodesInitial(blk,aoa1,aoa2,piv_pt)
+        SUBROUTINE shiftSurfaceNodesInitial(blk,aoa, piv_pt, ita, &
+             deltat, dxmin, totime)
```
tomneep:
```suggestion
        SUBROUTINE shiftSurfaceNodesInitial(blk, aoa, piv_pt, ita, deltat, dxmin, totime)
```

tomneep (approved):
Two very minor nitpicks that are somewhat optional. Approving now, so can be merged as and when you want.

PR #287, opened 2026-01-13
Attempt to add CI for ford

tomneep:
Builds the documentation using ford and uploads the resulting `doc` folder as an artifact.


mcbarton (approved):
The documentation has been added as an artifact. Once the repo has been made public the artifact could be used to make Github page.

The only change I would make, but is upto you, would be to remove any reference to main in the workflow yml file, since we no longer plan to merge dev into main.

PR #299, opened 2026-01-14
Make omega a member of the Block_t type

### LINKED ISSUE #295: Setting `omega` values
Currently, 4 values of `omega` can be set in the input file. These are used in the poisson solver with `omega1` used for block 1, `omega2` for block 2 and `omega3` for block 3. Beyond that every block uses the same values of `omega = omega4`. We can keep this behaviour by making `omega` a variable of the `Block_t` type and setting it when we read the input file. 

Whether or not we'd ever want finer control of the `omega` values e.g. to set a different value for block 100, is a different issue I think

tomneep:
Rather than having four variables we need to carry around as `omega` values, we instead make `omega` a member of `Block_t` which I think makes most sense.

Closes #295 (perhaps need to open an issue to discuss how we'd set higher omega values if desired)

### DIFF
--- a/src/ReadInput.f90
+++ b/src/ReadInput.f90
@@ -22,15 +22,13 @@ module biocfd_read_input
       !> `butter_move.dat`, and `block_details.dat`. From these the
       !> "grid file" names are constructed and those files are then
       !> read to determine the grid.
-      SUBROUTINE readInput(surGeoPoints,char_f,istart,itamax,pcItaMax,aoa,phase_angle,piv_pt, &
-                           mu_f, rho_f, inor, deltat,dxmin, &
-                           epsi,freq,omega1,omega2,omega3,omega4, re, uc)
+      SUBROUTINE readInput(surGeoPoints, char_f, istart, itamax, pcItaMax, aoa, phase_angle, &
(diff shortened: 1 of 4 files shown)


mcbarton:
@tomneep is this PR ready for review, or do you still have to test it?

tomneep:
I've tested locally and it gives consistent results. If you could look carefully though to double check I've not done anything wrong that would be great.

src/ReadInput.f90
```diff
@@ -41,6 +39,9 @@ SUBROUTINE readInput(surGeoPoints,char_f,istart,itamax,pcItaMax,aoa,phase_angle,
        REAL(dp) :: alpha_m,theta_m,alpha_m1,theta_m1, l_c,u_tip,disp
        INTEGER (int64) :: intflines, nblocks
        real(dp) :: dt_order, u0
+       ! Omega variables are set as block variables
```
mcbarton, line 42:
```suggestion
       !> Omega variables are set as block variables
```

tomneep: This will all change with #300 so I'd rather leave the change until then

mcbarton (approved):

PR #300, opened 2026-01-14
Start to document namelist

tomneep:
Putting as draft to save CI until I've done a pass through


tomneep:
This is probably all I feel confident documenting for now. Should be a starting point for others.

mcbarton (approved):

PR #301, opened 2026-01-15
Begin adding docstrings to StressCalc module

Arjun-Shergill:
Testing the docstrings to see if they appear in ford-docs

### DIFF
--- a/src/StressCalc.f90
+++ b/src/StressCalc.f90
@@ -1,32 +1,43 @@
+!> This is the module that computes drag and lift coefficients on the immersed body
 module biocfd_stress_calculation
   use, intrinsic :: iso_fortran_env, only: dp => real64, int64
   use biocfd_forcing, only: compute_value_and_derivatives
   use biocfd_block_type, only: Block_t
   IMPLICIT NONE
-
+!> Everything is private apart from the subroutine stress_calculation (exposed to other modules)
   private
 
   public :: stress_calculation
 
 contains
-
+!> This is the subroutine that computes viscous and pressure forces on IB surfaces
   SUBROUTINE stress_calculation(blk, mu_f, rho_f)
     !> The block to perform the stress calculation on (not this is
     !> inout only because of setting thetaDot and thetaDDot, which I
     !> think might not be needed)
     type(Block_t), intent(inout) :: blk
     !> Parameters used in the stress calculation
+    !> Fluid dynamic viscosity and Density
     real(dp), intent(in) :: mu_f, rho_f
-
+    !> ielem: Immersed Boundary element
+    !> i_x1, i_y1, i_z1: Where to interpolate fluid values
     INTEGER :: ielem, i_x1, i_y1, i_z1
-
+    !> i_cell, j_cell, j_cell: Which grid cell the surface point lies inside
     INTEGER :: i_cell, j_cell, k_cell
-
+    !> Diagis: Diagonal distance of the cell, normdis is the distance from the surface to the fluid sample point
     REAL(dp) :: diagdis, normdis, aval, bval, cval, del_X, del_Y, del_Z
-
+    !>pos1_x, pos1_y, pos1_z: Fluid sample point, point that is slightly off the surface, inside the fluid.
+    !> p_pos1: Pressure at this fluid sample point.
+    !> dpdn: Pressure gradient at the surface
+    !> dpdn_e: Pressure gradient near the wall, slightly inside the fluid.
+    !> usurf, vsurf, wsurf: Velocity at the surface
+    !> u_pos1, v_pos1, w_pos1: Fluid velocity at the fluid sample point
+    !>dudn_e, dvdn_e, dwdn_e: Velocity derivatives near the wall, slightly inside the fluid.
+    !> ddn_s: Tangential velocity gradient at wall
     REAL(dp) :: pos1_x, pos1_y, pos1_z, p_pos1, dpdn, dpdn_e, usurf, u_pos1, vsurf, v_pos1, &
                wsurf, w_pos1,  dudn_e, dvdn_e, dwdn_e, ddn_s(3)
-
+    !> alen: Element length
+    !>
     REAL(dp) :: alen, area, area_xz, area_yz, area_xy
 
     REAL(dp) :: shear_force(3), f_surf(3)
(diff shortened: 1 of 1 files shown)


src/StressCalc.f90
```diff
@@ -1,45 +1,57 @@
+!> This is the module that computes drag and lift coefficients on the immersed body
 module biocfd_stress_calculation
   use, intrinsic :: iso_fortran_env, only: dp => real64, int64
-  use biocfd_forcing, only: compute_value_and_derivatives, find_index_in_array
+  use biocfd_forcing, only: compute_value_and_derivatives
   use biocfd_block_type, only: Block_t
   IMPLICIT NONE
-
+!> Everything is private apart from the subroutine stress_calculation (exposed to other modules)
```
tomneep:
```suggestion
! Everything is private apart from the subroutine stress_calculation (exposed to other modules)
```

This is just a comment, not a docstring, so just use a plain `!`. I'm also not sure it is needed?

src/StressCalc.f90
```diff
@@ -1,45 +1,57 @@
...
   private
 
   public :: stress_calculation
 
 contains
-
+!> This is the subroutine that computes viscous and pressure forces on IB surfaces
```
tomneep:
```suggestion
!> Compute viscous and pressure forces on IB surfaces
```

No need to say "This is a subroutine..."

src/StressCalc.f90
```diff
@@ -1,45 +1,57 @@
...
   SUBROUTINE stress_calculation(blk, mu_f, rho_f)
     !> The block to perform the stress calculation on (not this is
     !> inout only because of setting thetaDot and thetaDDot, which I
     !> think might not be needed)
     type(Block_t), intent(inout) :: blk
     !> Parameters used in the stress calculation
+    !> Fluid dynamic viscosity and Density
```
tomneep:
```suggestion
    !> Fluid dynamic viscosity and Density
```

Remove the "Parameters used in the stress calculation" line since it shouldn't really be here

src/StressCalc.f90
```diff
@@ -1,45 +1,57 @@
+!> This is the module that computes drag and lift coefficients on the immersed body
```
tomneep:
```suggestion
!> This module computes drag and lift coefficients on the immersed body
```

src/StressCalc.f90
```diff
@@ -1,45 +1,57 @@
...
+    !> alen: Element length
+    !>
     REAL(dp) :: alen, area, area_xz, area_yz, area_xy
 
     REAL(dp) :: shear_force(3), f_surf(3)
 
     REAL(dp) :: pressureDrag, viscousDrag, viscousLift, pressureLift, area_Sx, area_Sy, surf_area
 
     real(dp) :: ac_y, ac_z, at_y, at_z
+    !> Note: The (3) at the end of some variables indicates that it is a 1D array with three components
```
tomneep:
```suggestion
    ! Note: The (3) at the end of some variables indicates that it is a 1D array with three components
```
This is a comment, not a docstring so should just be a plain `!`. I'd actually suggest deleting this line all together since this happens throughout the code and it seems a bit pointless.

src/StressCalc.f90
```diff
@@ -1,45 +1,57 @@
...
     real(dp) :: derivatives(3)
     ! In this case I think it makes sense to store the normal - there
     ! is an argument to make cosAlpha, cosBeta, and cosGamma a length
     ! 3 array in Block_t
     real(dp) :: normal(3)
-
+    !> Set all quantities to zero before looping
```
tomneep:
```suggestion
    ! Set all quantities to zero before looping
```

src/StressCalc.f90
```diff
@@ -62,8 +74,9 @@ SUBROUTINE stress_calculation(blk, mu_f, rho_f)
     !$acc firstprivate (rho_f, mu_f) &
     !$acc private(derivatives, normal) &
     !$acc reduction(+: pressureDrag, viscousDrag, viscousLift, pressureLift, area_Sx, area_Sy, surf_area)
+    !> Loop over all immersed boundary surface elements
```
tomneep:
```suggestion
    ! Loop over all immersed boundary surface elements
```

src/StressCalc.f90
```diff
@@ -74,15 +87,15 @@ SUBROUTINE stress_calculation(blk, mu_f, rho_f)
        del_X = blk%x1(i_cell+1)-blk%x1(i_cell)
        del_Y = blk%y1(j_cell+1)-blk%y1(j_cell)
        del_Z = blk%z1(k_cell+1)-blk%z1(k_cell)
-
+       !> This is the distance to the nearest interpolation point - It is the length of the cell
```
tomneep:
```suggestion
       ! This is the distance to the nearest interpolation point - It is the length of the cell
```

src/StressCalc.f90
```diff
@@ -93,11 +106,11 @@ SUBROUTINE stress_calculation(blk, mu_f, rho_f)
           blk% thetaDot  = blk% thetaDot2
           blk% thetaDDot = blk% thetaDDot2
        END IF
-
+       !> This is the main rigid body kinemtaics
```
tomneep:
```suggestion
       ! This is the main rigid body kinemtaics
```

src/StressCalc.f90
```diff
@@ -93,11 +106,11 @@ SUBROUTINE stress_calculation(blk, mu_f, rho_f)
...
        usurf = blk%xdot
        vsurf = -blk%thetaDot * (blk%zcent(ielem) - blk%piv_z) + blk%ydot
        wsurf = blk%thetaDot * (blk%ycent(ielem) - blk%piv_y)
-
+       !> Calculating Centripetal and tangential acceleration
```
tomneep:
```suggestion
       ! Calculating Centripetal and tangential acceleration
```

tomneep (changes requested):
Hi Arjun,

Thanks for this. I've made a few suggestions. The main thing is that we want to distinguish regular comments (`!`) from docstrings (`!>`). Unless something is going to appear in the auto-generated documentation it should just be a plain comment. 

There are some places where you documented internal variables of the subroutine that arguably should be regular comments.

The other thing to mention is that it is usually convention to keep comment line widths to something like 70-80 characters. The comments here are fine, but bear in mind you don't want to have really long comments without line breaks.

tomneep (approved):

PR #303, opened 2026-01-19
Bug fix pressure forcing index

### LINKED ISSUE #278: Indexing in `pressureForcingGhost`
Hi @SiddharthDS, 

I have a quick question about this line:

https://github.com/BioFSILab/Bio-CFD/blob/ba49203ad1f61a2c9ca278e20ddd47d2e0db85e6/src/subForcing.f90#L863C184-L863C223

Specifically, the final `-block(g)%cosBeta(block(g)%index_ts(n))` part. Everywhere else in this subroutine we seem to index with `block(g)%nelp(block(g)%index_ts(n))` i.e. with the `nelp` part, but here we just use a bare `block(g)%index_ts(n)`. Is this intentional? It may well be but I just wanted to double check!

tomneep:
Closes #278. 

Could probably do with a bit more testing, but this fixes the issue while adding a little tidy up. Will try and get it tested/merged while @mcbarton and I are still on the project.

### DIFF
--- a/src/Forcing.f90
+++ b/src/Forcing.f90
@@ -365,50 +365,53 @@ SUBROUTINE pressureForcingGhost(blk)
       REAL (dp) :: n1, pos1_x, pos1_y, pos1_z, pt1, &
                          aval, bval, cval, p_pos1, sur2nodeDis, dpdn, &
                          dpdn_e, ac_y, ac_z, at_y, at_z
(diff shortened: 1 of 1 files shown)


tomneep:
I'm closing this in favour of #310. Something like this PR could come later, along with a larger refactoring of this part of Forcing.

PR #305, opened 2026-01-19
Remove [xyz]pos variables

tomneep:
As far as I can tell, we aren't using `[xyz]pos` anywhere, so this just removes them. This is the `grep` output

```shell
❯ grep -iwR "[xyz]pos" src app
src/Blocks.f90:     REAL(dp) :: ymove,ypos
src/Blocks.f90:     !zpos not used
src/Blocks.f90:     REAL (dp) :: xmove,xpos, zmove,zpos
src/Search.F90:        blk%ypos=blk%yshift
src/Search.F90:        blk%xpos=blk%xshift
src/Search.F90:        blk%ypos =  blk%ypos  + blk%ychg
src/Search.F90:        blk%xpos =  blk%xpos  + blk%xmove
```


mcbarton (approved):
LGTM. Feel free to take the PR in once the ci passes all jobs.

PR #314, opened 2026-03-13
Add text to README.md to acknowledge support from Baskerville funding

warrickball:

PR #315, opened 2026-03-13
Set a maximum version for fortitude due to new rules

tomneep:
New rules added in v0.8.0 cause the CI to fail, so pin the maximum version. Fixing them won't take too long but is more time than I have. 

#316 is the issue for fixing the new rules in v0.8.0.


tomneep:
@warrickball this sets a maximum fortitude version so that the CI passes.

warrickball (approved):
LGTM. Linting CI passing here implies it fixed the failure we saw in #314.

